Issue #16, opened 2022-12-22
language server: code completetion


dalance:
For example, completing module instantiation with port.

PR #156, opened 2023-02-28
Add toplevel entity and keyword completion

dalance:
Add `module`, `interface`, `package`, and keywords completion support to veryl-ls


dalance:
Closes #16

PR #207, opened 2023-03-25
Add hash delay for always combination.

YSawc:
Add hash delay statement for always combination for case such as clocking test with embedding logic analizer.
I'm beginner of HDL, so if missing expression in compiled system verilog file, can you tell me about?


dalance:
Currrently, simulation only description is not allowed intendedly.
I'll consider how these description should be described at #209.

PR #326, opened 2023-07-30
Typedef

nblei:
Added support for typedefs using rust-like syntax:

```
     type word_t = logic<16>;
     type words_t = logic<16,16>;
     type regfile_t = word_t [16];
     type octbyte = bit<8> [8];
```


crates/analyzer/src/handlers/create_symbol_table.rs
```diff
@@ -222,6 +222,21 @@ impl<'a> VerylGrammarTrait for CreateSymbolTable<'a> {
...
 
+    fn type_def_declaration(&mut self, arg: &TypeDefDeclaration) -> Result<(), ParolError> {
+        match self.point {
+            HandlerPoint::Before => {
+                let r#type = arg.array_type.as_ref().into();
+                let property = TypeDefProperty { r#type };
+                let kind = SymbolKind::TypeDef(property);
+                self.insert_symbol(&arg.identifier.identifier_token, kind);
+                let name = arg.identifier.identifier_token.token.text;
+                self.namespace.push(name)
```
dalance, line 233: `namespace.push` and `pop` is not required because `type` doesn't construct inner namespace.

crates/emitter/src/emitter.rs
```diff
@@ -333,6 +339,7 @@ impl Emitter {
                 }
             }
             dbg!("a");
+            dbg!(arg);
```
dalance, line 342: Sorry. The above `dbg` is mistake.
Could you remove all `dbg`?

crates/emitter/src/emitter.rs
```diff
@@ -1259,6 +1266,21 @@ impl VerylWalker for Emitter {
         self.semicolon(&arg.semicolon);
     }
 
+    /// Semantic action for non-terminal 'LocalparamDeclaration'
+    fn type_def_declaration(&mut self, arg: &TypeDefDeclaration) {
+        self.str("typedef");
```
dalance, line 1271: The following code is better. This is because the original token holds comments and spacing around the token.

`self.token(&arg.r#type.type_token.replace("typedef"));`

dalance (commented):
Looks good for the syntax.
I added some comment about implementation.
Could you add formatter implementation too?

crates/emitter/src/emitter.rs
nblei: My mistake. Fixed.

crates/emitter/src/emitter.rs
nblei: Fixed.

crates/analyzer/src/handlers/create_symbol_table.rs
nblei: Fixed.

book/src/04_language_reference/03_data_type/02_user_defined_type.md
```diff
@@ -43,3 +43,13 @@ module A {
     assign a = EnumA::member_a;
 }
 ```
+
+## Typedef
+
+The `type` keyword can be used to define a typedef to scalar or array types.
+
+```veryl,playground
+type word_t = logic<16>;
```
dalance, line 52: Could you add `module A {}` to pass playground compile?

dalance:
All changes look good.

PR #327, opened 2023-07-30
No longer prefix verilog package identifiers with veryl project names

nblei:
This is in order to deal with the issue I raised https://github.com/dalance/veryl/issues/324

I am not sure entirely why module and package names are prefixed with project names --- perhaps because verilog has no real module system or namespace system.  If that's the goal, then `emitter.rs` should be changed to also prefix package identifiers in `scoped_identifier`.


dalance:
I intended that the prefix is used to realize package namespace.
So `scoped_identifier` should be prefixed by package identifier.

nblei:
OK Dalance, I will withdraw this pull request because it seems like you have a better solution in mind.

PR #357, opened 2023-08-17
Added union type

nblei:
Added a union type which transpiles to a SystemVerilog `packed union`.

I also changed the parser's `StructDeclaration` and sub components to `StructUnionDeclaration`, etc, as they have the same syntactic structure.

I would like to add more rigorous type checking on unions.  Currently, we do not enforce that all variants have the same width.

This will entail a larger effort to add more type checking in general to Veryl's transpiler.


dalance:
Looks good.
I found a formatting issue, but it seems to be caused by original struct formatting issue.
I'll fix it after merging this PR.

```
    typedef  <- a strange new line
    union packed {
        logic   b;
        Boolean c;
    } A;
```

PR #383, opened 2023-09-06
Cycle check

nblei:
This adds type checking for cyclic dependencies:

image

These dependencies are checked for `type` aliases, `struct`s and `union`s.

This is done with `CreateTypeDag` handler in Pass2.  It runs in Pass2 because it requires the symbol table made in Pass1.

Follow on work will be to add further type checking on type sizes.  This will go into Pass3, as it requires the [TypeDag] generated in Pass2.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -628,6 +628,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "daggy"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


nblei:
Now also checks for undefined types:

image

Also, the Type DAG now stores sufficient information to create precise `AnalyzerError`s when traversing the DAG, rather than while traversing the AST.

dalance:
Looks good!

There are some commented-out functions.
Can I merge them as is?

nblei:
I pushed a commit removing those commented out functions.

dalance:
Thanks.
I'll merge this PR.

PR #392, opened 2023-09-20
Migrate to parol v0.24.0

jsinger67:
Changes necessary for update to parol 0.24.0
Background: Genrated parsers implement error recovery startegies. This had influence on some error handling and reporting


dalance:
Thanks!

PR #393, opened 2023-09-22
Checks that (module) inputs are only read

nblei:
This commit checks that module inputs are read-only.

Attempts to assign a value to an input, either via an assignment statement or in a procedural block, will result in an `assignment_to_input` error.


dalance:
Looks good. I'll merge this PR.
Thanks!

PR #493, opened 2024-02-06
support void type of function

YSawc:
modify support void type of function. I noticed that only reference assignment function that does not returns value is not supported yet.


dalance:
I think there is another solution to support void function.

```
// void return type support
function Func () -> void {
}

// omittable return type
function Func () {
}
```

How do you think?

YSawc:
It sounds good! May i change emitter to apply your suggestion?

YSawc:
@dalance 
[cecf6b7](https://github.com/dalance/veryl/pull/493/commits/cecf6b7abee8d2815735438e4d746e650eff1954)
I changed to apply omit return type. Please confirm changed logic.

crates/emitter/src/aligner.rs
```diff
@@ -776,7 +776,10 @@ impl VerylWalker for Aligner {
         if let Some(ref x) = arg.function_declaration_opt0 {
             self.port_declaration(&x.port_declaration);
         }
-        self.minus_g_t(&arg.minus_g_t);
+        if let Some(ref x) = arg.function_declaration_opt1 {
+            self.minus_g_t(&x.minus_g_t);
+            self.scalar_type(&x.scalar_type);
```
dalance, line 781:
```suggestion
            // skip type align
            //self.scalar_type(&x.scalar_type);
```

crates/emitter/src/aligner.rs
```diff
@@ -776,7 +776,10 @@ impl VerylWalker for Aligner {
...
+        }
         // skip type align
         //self.scalar_type(&arg.scalar_type);
```
dalance, line 784:
```suggestion
```

crates/emitter/src/aligner.rs
dalance: The `scalar_type` should be skipped.
This causes the following unexpected spaces.

```
function automatic logic        [ParamX-1:0] FuncA(
                         ^^^^^^^
        input  logic        [ParamX-1:0] a,
                     ^^^^^^^
```

dalance (commented):
Looks good!
I've added a comment about alignment of emitter.
Could you check it?

YSawc:
Thanks your feedback! I applied each suggestions.

dalance:
Thanks!!
I'll merge this PR.

PR #585, opened 2024-03-21
Add strip_comments option

### LINKED ISSUE #580: Generate SV code without comments
The ability to generate documentation from the veryl code is extremely useful. 

It would also be helpful to have an option to remove or strip away the comments that are in the veryl code when generating the SystemVerilog code. In cases where the veryl code is heavily documented, this would allow inspection of the SystemVerilog code in a plain format and make it easier to review the generated code.

This process to do this would simply be placing a "strip_comments"  in the toml / veryl file as: 

```toml 
[format]
indent_width = 4
strip_comments = true 

```

dalance:
Closes #580

### DIFF
--- a/crates/emitter/src/emitter.rs
+++ b/crates/emitter/src/emitter.rs
@@ -191,7 +191,8 @@ impl Emitter {
             self.space(*width as usize);
         }
 
-        if duplicated.is_some() {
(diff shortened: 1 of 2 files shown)


saturn77:
A fast update, great work.

PR #605, opened 2024-03-27
add 'case' keyword

taichi-ishitani:
Currently, VS code plugin does not highlight `case` keyword.
This PR is to fix this problem.


dalance:
Looks good.
Thanks!

Issue #569, opened 2024-03-18
[Feature] Omit specifying clock and reset signals


taichi-ishitani:
Most modules have one pair of clock and reset only so we can omit spacyfying clock and reset singals on `always_ff` blocks.

```systemverilog
always_ff {
  if_reset {
    q = 0;
  } else {
    q = i_d;
  }
}
```

To realize this feature, special types named `clock` and `reset` would be instroduced. These types are to tell Veryl which signals are clock and reset signals.
If a module have only one pair of signals of which types are `clock` and `reset` then Veryl can insert that clock and reset signals to `always_ff` blocks automatically.

```systemverilog
module foo (
  i_clk:   input  clock,
  i_rst_n: input  reset,
  i_d:     input  logic,
  o_q:     output logic
) {
  var q: logic;

  assign o_q = q;
  always_ff { // Veryl can convert this code to 'always_ff (i_clk, i_rst_n)'
    if_reset {
      q = 0;
    } else {
      q = i_d;
    }
  }
}
```

```systemverilog
module foo (
  i_clk_a:   input  clock,
  i_rst_a_n: input  reset,
  i_clk_b:   input  clock,
  i_rst_b_n: input  reset,
  i_d:       input  logic,
  o_q:       output logic
) {
  var q: logic;

  assign o_q = q;
  always_ff { // Veryl doesn't know which clock/reset pairs should be used. This code causes an error.
    if_reset {
      q = 0;
    } else {
      q = i_d;
    }
  }
}
```

```systemverilog
module foo (
  i_clk:   input  clock<4>,
  i_rst_n: input  reset<4>,
  i_d:     input  logic<4>,
  o_q:     output logic<4>
) {
  for i in 0..4 :g {
    let clk:   clock = i_clk[i];
    let rst_n: reset = i_rst_n[i];
    var q:     logic;

    assign o_q[i] = q;
    always_ff { // block 'g' has one clock/reset pair only so Veryl can convert this to 'always_ff (clk, rst_n)'
      if_reset {
        q = 0;
      } else {
        q = i_d[i];
      }
    }
  }

}
```

In addition, I think synthesis tools require that clock and reset signals should be single bit signals.
Therefore, Veryl should check if signals speciled as clock/reset are single bit signals.

saturn77:
I think this is a good feature that will make the code more compact. 

My suggestion is to allow for naming the actual clock and reset signals in the toml file, such as 

```toml 
[build]
clock_name = "clock"
[...]

PR #714, opened 2024-05-10
allow to omit specifying clock and reset signals

taichi-ishitani:
resolve veryl-lang/veryl#569


crates/analyzer/src/handlers/create_symbol_table.rs
```diff
@@ -197,7 +272,14 @@ impl<'a> VerylGrammarTrait for CreateSymbolTable<'a> {
                 affiniation,
             };
             let kind = SymbolKind::Variable(property);
-            self.insert_symbol(&arg.identifier.identifier_token.token, kind, false);
+            let id =
+                self.insert_symbol(&arg.identifier.identifier_token.token, kind.clone(), false);
+
+            if self.is_default_clock_candidate(kind.clone()) {
+                self.default_clock_candidates.push(id.unwrap());
```
dalance, line 279: Could you handle the case of `id` is `None`?
Probably "duplicated identifier" causes panic at `id.unwrap()`.

taichi-ishitani: My understanding is that "duplicated identidifier" will be checked here.
https://github.com/veryl-lang/veryl/blob/c2fcbe9d35cd7e839eb79cefeb4ad2e2fba843fe/crates/analyzer/src/handlers/create_symbol_table.rs#L96-L103

Therefore, I think we can handle `id` as a `Some` object.

dalance: If `id` is `None`, the error are registerd by `self.errors.push` and the `id` is returned normally.
So we can't assume `id` is always `Some`.

taichi-ishitani: I understand.
I will fix it.

taichi-ishitani: I've fixed your feedback.

dalance:
Looks good.
Thanks!

Issue #712, opened 2024-05-10
SystemVerilog keyword check


dalance:
Some keyword of SystemVerilog can be used as identifier in Veryl.
They will cause syntax error of transpiled SystemVerilog code.
So they should be checked by Veryl compiler.

taichi-ishitani:
Keywords list
(From IEEE 1800-2023 Annex B)

* accept_on
* alias
* always
* always_comb
* always_ff
* always_latch
* and
* assert
* assign
* assume
* automatic
* before
* begin
* bind
* bins
* binsof
* bit
* break
* buf
* bufif0
* bufif1
* byte
* case
* casex
* casez
* cell
* chandle
* checker
* class
* clocking
* cmos
* config
* const
* constraint
* context
* continue
* cover
* covergroup
* coverpoint
* cross
* deassign
* default
* defparam
* design
* disable
* dist
* do
* edge
* else
* end
* endcase
* endchecker
* endclass
* endclocking
* endconfig
* endfunction
* endgenerate
* endgroup
* endinterface
* endmodule
* endpackage
* endprimitive
* endprogram
* endproperty
* endspecify
* endsequence
* endtable
* endtask
* enum
* event
* eventually
* expect
* export
* extends
* extern
* final
* first_match
* for
* force
* foreach
* forever
* fork
* forkjoin
* function
* generate
* genvar
* global
* highz0
* highz1
* if
* iff
* ifnone
* ignore_bins
* illegal_bins
* implements
* implies
* import
* incdir
* include
* initial
* inout
* input
* inside
* instance
* int
* integer
* interconnect
* interface
* intersect
* join
* join_any
* join_none
* large
* let
* liblist
* library
* local
* localparam
* logic
* longint
* macromodule
* matches
* medium
* modport
* module
* nand
* negedge
* nettype
* new
* nexttime
* nmos
* nor
* noshowcancelled
* not
* notif0
* notif1
* null
* or
* output
* package
* packed
* parameter
* pmos
* posedge
* primitive
* priority
* program
* property
* protected
* pull0
* pull1
* pulldown
* pullup
* pulsestyle_ondetect
* pulsestyle_onevent
* pure
* rand
* randc
* randcase
* randsequence
* rcmos
* real
* realtime
* ref
* reg
* reject_on
* release
* repeat
* restrict
* return
* rnmos
* rpmos
* rtran
* rtranif0
* rtranif1
* s_always
* s_eventually
* s_nexttime
* s_until
[...]

PR #750, opened 2024-06-03
Identifies SV keywords used as identifiers

nblei:
For issue #712

I added a `var_declaration` definition for `CheckIdentifier`.  In it, I do not call `check`, but rather just do the work in the `var_declaration` function itself --- perhaps this should be merged into `check`?


dalance:
> perhaps this should be merged into check?

Yes. `check` is called at arbitrary identifier definition not limited to `var_declaration`.
So if the `is_sv_keyword` is merged into `check`, other identifier like param, function, and so on will be checked too.

nblei:
> > perhaps this should be merged into check?
> 
> Yes. `check` is called at arbitrary identifier definition not limited to `var_declaration`. So if the `is_sv_keyword` is merged into `check`, other identifier like param, function, and so on will be checked too.

OK, I'm just running the SV keyword checks in `check` now.

dalance:
Could you add test for `SvKeywordUsage` error at `crates/analyzer/src/tests.rs`?

nblei:
> Could you add test for `SvKeywordUsage` error at `crates/analyzer/src/tests.rs`?

Done.

Issue #746, opened 2024-05-31
Function call for non functoin identifier


taichi-ishitani:
Veryl does not report an error for a function call for non function indeintifier.

![image](https://github.com/veryl-lang/veryl/assets/2922232/9db80f09-0e95-410f-8567-f51510d73577)

PR #753, opened 2024-06-04
Function identifier as factor fix

### LINKED ISSUE #747: Function identifier refered as a factor
Veryl does not report an error for a function identifier refered as a factor like below.
image

nblei:
I created a `CheckExpression` pass.   Currently, it only looks at the kinds  of factors, and raises an error when the kind is not appropriate (e.g., functions, modules, interfaces, etc.), and it also checks to ensure only function kinds (`Function`, `SystemVerilog`, `ModportFunctionMember`, `SystemFunction`) are function call factors.

So this fixes #747  and #746.


dalance:
Looks good.
I'll merge this PR after resolving the conflict.

Issue #749, opened 2024-06-03
Generics arity error inside generic package


taichi-ishitani:
Veryl reports a generic arity error for the code below.

```systemverilog
package PackageA::<W> {
  struct StructA {
    a: logic<W>
  }

  function FuncA() -> StructA {
    var a: StructA;
    a.a = 0;
    return a;
  }
}
```

```
https://doc.veryl-lang.org/book/06_appendix/02_semantic_error.html#mismatch_generics_arity\mismatch_generics_arity (link)]8;;\

  💥 generics "StructA" has 1 generic arguments, but 0 arguments are supplied
   ╭─[:5:1]
 5 │ 
 6 │   function FuncA() -> StructA {
   ·                       ───┬───
   ·                          ╰── Error location
 7 │     var a: StructA;
   ╰────
  help: fix generics arguments
```

I think Veryl should not report this error because the `StructA` struct is defined within the same package scope.

taichi-ishitani:
Generated SV file is also strange.

```systemverilog
package veryl_PackageA;
    typedef struct packed {
        logic [-1:0] a;
    } __StructA__W;

    function automatic __StructA__W FuncA() ;
        __StructA__W a  ;

        a.a = 0;
        return a;
    endfunction
endpackage

module veryl_ModuleA #(
    parameter int unsigned W = 1
);
    veryl_PackageA::__StructA__W a  ;
    always_comb a.a = 0;
endmodule
```

taichi-ishitani:
Package and struct can have generic parameters but Veryl doent not support this pattern.
https://github.com/veryl-lang/veryl/blob/365ea2e920acc63a956cae2f3b00226b1372f180/crates/analyzer/src/handlers/create_symbol_table.rs#L48

Due to this, the generics identidifier error was raised.

taichi-ishitani:
To support hierarchical generics reference, need to introduce namespace for generics parameters?

PR #755, opened 2024-06-05
Support hierarchical generic parameter reference

taichi-ishitani:
Currently Veryl does not support hierarchical generic parameter reference.
Because of this, #749 occurs.

This PR is to add this feature.

fix veryl-lang/veryl#749


taichi-ishitani:
I've fixed your feedbacks.

PR #766, opened 2024-06-08
Reports error on assignment to const value declared by 'let' statement

nblei:
```systemverilog
module top (
    i_clk: input clock,
    i_rst: input reset,
) {
    always_comb {
        let y: logic = 1;
        y = 0;
    }
}
```

Previously, this would not report any errors.


dalance:
Looks good!
I'll merge this PR.

PR #771, opened 2024-06-10
Enum reset x

nblei:
Allows `EnumMember`s with 'x' or 'z' values to be `Factor`s in assignment operations inside `if_reset` blocks.

 Previously,

```systemverilog
module top (
    i_clk: input clock,
    i_rst: input reset,
) {

    enum State: logic<3> {
        Idle = 3'bxx1,
        Run0 = 3'b000,
        Run1 = 3'b010,
        Run2 = 3'b100,
        Done = 3'b110,
    }

    var state: State;

    always_ff {
        if_reset {
            state = State::Idle;
        }
    }

}
```

would result in an error because of the don't care bits in `State::Idle`.


nblei:
@dalance 

One thing that is strange about this is that `EnumItem`s are not `evaluate`d.  When are symbols supposed to be evaluated?

dalance:
Evaluation is executed at symbol table resolve.
https://github.com/veryl-lang/veryl/blob/68c3593a1c5a53bbc2fdfdae4173270b048b7513/crates/analyzer/src/symbol_table.rs#L173

But it may be unnecessary. I think calling `symbol.found.evaluate()` is sufficient when it is necessary.
If the symbol is already evalated, it returns the cached `evaluated`, and if not, it evaluates at the call time.

Therefore, the code of `evaluator.rs` can be simplified like below:

```rust
// from
if let Some(evaluated) = symbol.found.evaluated.get() {
    evaluated
} else {
    symbol.found.evaluate()
}

// to
symbol.found.evaluate()
```

taichi-ishitani:
For SystemVerilog, there is the limitation below for unassigned enum labels following the enum label with x/z value.

From IEEE 1800-2023 6.19.2 Enumerated type ranges
image

I think you need to check if this limitation is satisfied.

dalance:
@nblei 
I think this PR can be merged.
Is it ok with you?

@taichi-ishitani 
I think this change is about evaluating enum value, 
and your suggestion seems to be about requirement of additional check for enum declaration.
So could you create a new issue for it?

taichi-ishitani:
> @taichi-ishitani I think this change is about evaluating enum value, and your suggestion seems to be about requirement of additional check for enum declaration. So could you create a new issue for it?

Sure, I will create a new ticket.

nblei:
> @nblei I think this PR can be merged. Is it ok with you?

Yes.

Issue #775, opened 2024-06-10
Cannot use a generic parameter as a variable type


taichi-ishitani:
Veryl reports an error while building the code below.

```systemverilog
module ModuleA {
  function FuncA::<T> -> T {
    var a: T;
    a = 0;
    return a;
  }

  type my_logic = logic;
  let _a: logic = FuncA::<my_logic>();
}
```

```
]8;;https://doc.veryl-lang.org/book/06_appendix/02_semantic_error.html#mismatch_type\mismatch_type (link)]8;;\

  💥 "T" is expected to "enum or union or struct", but it is "generic
  │ parameter"
   ╭─[:2:1]
 2 │   function FuncA::<T> -> T {
 3 │     var a: T;
   ·         ┬
   ·         ╰── Error location
 4 │     a = 0;
   ╰────
```

taichi-ishitani:
This error occurs here.
https://github.com/veryl-lang/veryl/blob/5d16a2796c19cf8c7e314cd0865213a704b1d9a6/crates/analyzer/src/analyzer.rs#L104

Need to get the actual generic argument when that symbol is a generic parameter.

PR #778, opened 2024-06-11
Allow to use generic parameter as variable type

taichi-ishitani:
Currently, Veryl does not allow to use `generic parameter` as `variable type` so veryl-lang/veryl#775 occurs.
This PR is remove this limitation.

fix veryl-lang/veryl#775

### DIFF
--- a/crates/analyzer/src/symbol_path.rs
+++ b/crates/analyzer/src/symbol_path.rs
@@ -180,10 +180,34 @@ impl From<&syntax_tree::ExpressionIdentifier> for SymbolPathNamespace {
     }
 }
 
+#[derive(Copy, Debug, Clone, PartialEq, Eq)]
+pub enum GenericSymbolPathKind {
+    Identifier,
+    IntegerBased,
+    IntegerBaseLess,
+    IntegerAllBit,
+    RealExponent,
+    RealFixedPoint,
+}
+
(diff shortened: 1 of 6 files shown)


crates/analyzer/src/symbol_path.rs
```diff
@@ -180,10 +180,33 @@ impl From<&syntax_tree::ExpressionIdentifier> for SymbolPathNamespace {
...
+    Identifier,
+    IntegerBased,
+    IntegerBaseLess,
+    IntegerAllBit,
+    RealExponent,
+    RealFixedPoint,
+}
+
+impl GenericSymbolPathKind {
+    pub fn to_kind_name(&self) -> String {
```
dalance, line 194: How about implementing `Display` instead of `to_kind_name`.
Through `Display`, `to_string` is provided and stringify by `display!`, `format!` can be used.

```rust
impl fmt::Display for GenericSymbolPathKind {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        let mut text = String::new();
        ... // construct text
        text.fmt(f)
    }
}
```

taichi-ishitani: I implemented `fmt::Display` for `GenericSymbolPathKind `.

PR #786, opened 2024-06-12
fixed multiple issues with expression checking

nblei:
Fixes the issue raised by @nananapo in #785 but also fixes a number of other related issues.

1. Modports and interfaces were allowed to be in expressions if they were passed in to a module as a port.  This has been fixed.
2. While user defined functions are not higher ordered type operators, some SystemVerilog System Functions *are* --- namely, ones which extract size information from a type.  Thus, `Enum`, `Union`, `Struct`, `UnionMember`, and `StructMember` are suitable to be passed into functions with Kind `SystemFunction`.


crates/analyzer/src/handlers/check_expression.rs
```diff
@@ -36,17 +43,25 @@ impl<'a> VerylGrammarTrait for CheckExpression<'a> {
...
                             if x.factor_opt.is_none() {
-                                self.errors.push(AnalyzerError::invalid_factor(
-                                    &identifier,
-                                    &rr.found.kind.to_kind_name(),
-                                    self.text,
-                                    &token,
-                                ));
+                                self.errors.push(error);
+                            } else {
+                                self.call_stack_kind.push(FunctionKind::NonSystem);
```
dalance, line 57: I think `self.call_stack_kind.pop` corresponding to this push may be required at `HandlerPoint::After`.

nblei: You're correct.

crates/analyzer/src/handlers/check_expression.rs
```diff
@@ -93,6 +125,19 @@ impl<'a> VerylGrammarTrait for CheckExpression<'a> {
...
             }
+        } else if let Factor::ExpressionIdentifierFactorOpt(x) = arg {
+            let expid = x.expression_identifier.as_ref();
+            if let Ok(rr) = symbol_table::resolve(expid) {
+                match rr.found.kind {
+                    SymbolKind::Function(_) | SymbolKind::ModportFunctionMember(_) => {
+                        self.call_stack_kind.pop();
+                    }
+                    SymbolKind::SystemFunction => {
+                        self.call_stack_kind.push(FunctionKind::System);
```
dalance, line 136: Why is this `push` not `pop`?

nblei:
Because I'm an idiot

dalance:
Thanks.
I'll merge this PR after resolving the conflict.

Issue #861, opened 2024-08-08
[Feature] Expand`inside` expression and `case` statement


taichi-ishitani:
There are tools not supporting `inside` expression and `case inside` statement.
(refs #811)

For these tools, Veryl should have an option to expand `inside` expression and `case inside` statement.

taichi-ishitani:
It seems that Quartus and iverilog support the wildcard operator `==?`.
https://www.intel.com/content/www/us/en/programmable/quartushelp/17.0/mapIdTopics/jka1465580570693.htm
https://github.com/steveicarus/iverilog/blob/676b36e4554f119d0606ade79db2fba6834d3445/ivtest/ivltests/wild_cmp_net.v#L6

taichi-ishitani:
This `case` statement can be emitted like below.

```
    always_comb {
        case x {
            0: a = 1;
            1: a = 1;
            2: {
                   a = 1;
                   a = 1;
                   a = 1;
               }
            3, 4   : a = 1;
            5..=7  : a = 1;
            y - 1  : a = 1;
            default: a = 1;
        }
    }
```

```systemverilog
always_comb begin
  case (1'b1)
    x ==? 0: a = 1;
    x ==? 1: a = 1;
    x ==? 2: begin
      a = 1;
      a = 1;
      a = 1;
    end
    a ==? 3, a ==? 4:     a = 1;
    (a >= 5) && (a <= 7): a = 1;
    a ==? (y - 1):        a = 1;
    default:              a = 1;
   
  endcase
end
```

taichi-ishitani:
This `inside` oepration can be emitted:

```
assign a = inside 1 + 2 / 3 {0, 0..10, 1..=10};
```

```systemverilog
assign a = ((1 + 2 / 3) ==? 0) ||
           ((1 + 2 / 3) >= 0 && (1 + 2 / 3) < 10) ||
           ((1 + 2 / 3) >= 0 && (1 + 2 / 3) <= 10);
```

dalance:
Looks good.
I think both expansion is acceptable.

PR #873, opened 2024-08-10
Expand `inside` operation

taichi-ishitani:
resolve veryl-lang/veryl#861


crates/emitter/src/emitter.rs
```diff
@@ -317,11 +410,101 @@ impl Emitter {
...
+            self.str("(");
+            self.expression(lhs);
+            self.str(")");
+            self.space(1);
+            self.str(op_r);
+            self.space(1);
+            self.str("(");
+            self.expression(&x.expression);
+            self.str(")");
+            self.str(")");
```
dalance, line 497: This sequence seems to be bit long.
I think merging some `str` and `space` into a `str` is no problem. For example, `self.str(")) && ((")`.

taichi-ishitani: I fixed your feedback.

dalance:
Looks good.
Thanks!

Issue #933, opened 2024-09-03
Named block inside `generate` declaration 


taichi-ishitani:
SystemVerilog has no syntax for simple generate block like below.

```systemverilog
generate
  begin : g_foo
    logic foo;
  end
endgenerate
```

![image](https://github.com/user-attachments/assets/49ea91f4-13e0-4faf-9cd5-254db0e626c7)

So Veryl should emit any named blocks with `if (1)` header but a named block inside a `if`/`for` generate block does not start with `if (1)` header.

![image](https://github.com/user-attachments/assets/317dd671-ef51-4f5f-8b01-04a11f02de90)

PR #935, opened 2024-09-03
Fix wrong named block emittion

taichi-ishitani:
fix veryl-lang/veryl#933


dalance:

Issue #793, opened 2024-06-20
[Feature] Generics for code block


taichi-ishitani:
SV's text macro is useful feature to reduce coding cost for repeated code blocks.
I need this kind of feature to import code below to Very.
https://github.com/rggen/rggen-sample/blob/b0439abf7d5fa6086c4d0aef0ffcdc6836506f06/uart_csr.sv#L1-L9

I think it can be achieved if the generics feature handle code blocks.

dalance:
A syntax idea about generic `block` declaration.

```systemverilog
block rggen_connect_bit_field_if::<RIF, FIF, LSB, WIDTH> {
    assign  FIF.valid                 = RIF.valid;
    assign  FIF.read_mask             = RIF.read_mask[LSB+:WIDTH];
    assign  FIF.write_mask            = RIF.write_mask[LSB+:WIDTH];
    assign  FIF.write_data            = RIF.write_data[LSB+:WIDTH];
    assign  RIF.read_data[LSB+:WIDTH] = FIF.read_data;
    assign  RIF.value[LSB+:WIDTH]     = FIF.value;
}

module ModuleA {
    inst u_block: rggen_connect_bit_field_if::<bit_field_if, bit_field_sub_if, 0, 8>;
}
```

taichi-ishitani:
It looks good.
But why is instance name need? Is it to separate namespace?

dalance:
> It looks good. But why is instance name need? Is it to separate namespace?

Yes. And adding name-less instantiation increases syntax complexity.
The generated SV will become like below:

```systemverilog
if (1) begin: u_block
    assign  FIF.valid                 = RIF.valid;
    assign  FIF.read_mask             = RIF.read_mask[LSB+:WIDTH];
    assign  FIF.write_mask            = RIF.write_mask[LSB+:WIDTH];
    assign  FIF.write_data            = RIF.write_data[LSB+:WIDTH];
    assign  RIF.read_data[LSB+:WIDTH] = FIF.read_data;
    assign  RIF.value[LSB+:WIDTH]     = FIF.value;
end
```

taichi-ishitani:
Module instance and code block instance are completely different.
I think another syntax should be introduced for code block instnace.

taichi-ishitani:
How about `block` or `insert`?

```systemverilog
block u_block: rggen_connect_bit_field_if::<bit_field_if, bit_field_sub_if, 0, 8>;
[...]

PR #950, opened 2024-09-07
Merge `ModuleItem` and `InterfaceItem`

### LINKED ISSUE #908: Interface composition
If interface can have interface instances as the member, interface composition can be achived like below.
Syntax and use case should be more considerd.

```systemverilog
interface Read {
}

interface Write {
}

interface RW {
    inst read: Read;
    inst write: Write
}
```

Refs: https://www.reddit.com/r/FPGA/comments/1exhbtj/comment/ljdyafm/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

### LINKED ISSUE #947: [Feature] `always_*` blocks inside interface
SV interface can have `always_*` blocks.
Veryl should also support this feature.

refs: #908

taichi-ishitani:
For #793, generate constructions for module and interface need to be merged.
To do this, `GenerateItem` will be introduced and all items except for `ModportDeclaration` will be moved from `ModuleItem` and `InterfaceItem` to `GenerateItem`.

By merging `ModuleItem` and `InterfaceItem`, interface will be able to have interface instances (#908) and `always_*` blocks (#947).

resolve #947


dalance:
I think merging them is not so good because the items of them are not guaranteed the equivalence in the future even if they have the almost same items now.
If module only or interface only declaration is added in the future, additional semantic checking is required each time.

About #908, I think more specialized syntax may be required instead of mere interface instantiation.

taichi-ishitani:
This PR is also preparetion for #793.
I think #793 should be applied for both of module and interface but it is difficult to define the syntax for #793 without merging ModuleItem and ItnerfaceItem.

dalance:
I think #793 can be achieved by adding `BlockItem` individually.
But merging them at the moment and spliting them again in the future if syntax diverging occurs are acceptable too.
How do you think?

taichi-ishitani:
Module and interface have own syntax for generate constructs because ModuleItem and IterfaceItem are different.
I think we need to merge these `generate` syntax to introduce `BlockItem`. Currently, differences are

* Module cannot have modport.
* Interface has no sytnax for module/interface instantiation.

The 1st item can resolved by introducing a new limitation that modport cannot be declared in generate blocks. I think this is acceptable limitation.

The 2nd item can be resolved by introducing the existing sytnax for module instantiatoin to interface. But an additinal semantic check will be required like below.
https://github.com/veryl-lang/veryl/blob/9e5a7160b7d65ae52fb340ea05cf96972be6051e/crates/analyzer/src/handlers/check_type.rs#L231-L273

dalance:
OK. I agree merging the items.
I think `GenerateGroup` and `GenerateItem` may be better than `ModuleInterfaceGroup` and ` ModuleInterfaceItem` at the view point of naming consistency.

taichi-ishitani:
Thank you for your feedback.
I've renamed `ModuleItnerfaceItem` to `GenerateItem` and moved `ModportDelcaration` to `InterfaceItem`.

PR #976, opened 2024-09-29
Add `#[inline(never)]` to `expression*` as a workaround for long time compilation

### LINKED ISSUE #975: veryl-0.13.0 can't be built because of runaway builds
We have all amd64 builds failing because the Rust compiler doesn't produce an output, see the  [Ruust issue](https://github.com/rust-lang/rust/issues/130905).

Do you use any special options to build Veryl?

DianQK:
Closes #975.

In some build configurations, Rust/LLVM luckily meets the inlining threshold, which inlined all expression functions to get a huge function, causing compile time problems.

### DIFF
--- a/crates/emitter/src/emitter.rs
+++ b/crates/emitter/src/emitter.rs
@@ -1002,6 +1002,9 @@ impl VerylWalker for Emitter {
     }
 
     /// Semantic action for non-terminal 'Expression'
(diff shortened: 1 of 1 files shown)


dalance:
Thank you for your investigation!
I'll merge this PR, and add `#[inline(never)]` to other `expression`s in whole program to prevent another possible issue.

PR #1011, opened 2024-10-11
Execute `type check` afeter resolving generics

### LINKED ISSUE #1006: No error is reported for interface port with port direction
An interface port cannot have port direction so Veryl should report an error for this code but does not.

```
    port_i : input InterfaceA::<2>,
    modport_i : input InterfaceA::<3>
```

_Originally posted by @taichi-ishitani in https://github.com/veryl-lang/veryl/issues/1005#issuecomment-2403995006_

### LINKED ISSUE #1007: No error is reported for interface port without modport specification
An interface port should have modport specification but Veryl reports no error for an interface port without modport specification.

image

taichi-ishitani:
The `type check` checker requires resolved generic instanciations but, currently, it is executed before resolving them.
This is cause of #1006 and #1007.

Therefore, the checker should be executed after resolving generic instanciations.

fix #1006
fix #1007

### DIFF
--- a/crates/analyzer/src/handlers.rs
+++ b/crates/analyzer/src/handlers.rs
@@ -99,7 +99,6 @@ pub struct Pass2Handlers<'a> {
     check_enum: CheckEnum<'a>,
     check_modport: CheckModport<'a>,
     check_function: CheckFunction<'a>,
-    check_type: CheckType<'a>,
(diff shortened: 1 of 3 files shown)


dalance:
I understood it. I think this solution is right.

PR #1043, opened 2024-10-30
`cond_type` attribute

### LINKED ISSUE #937: Add `unique`, `unique0`, and `priority` keywords
Consider
```systemverilog
module test (
	input [2:0] onehot,
	output logic [1:0] idx
);
always_comb begin
	case (onehot)
		3'b001: idx = 2'b01;
		3'b010: idx = 2'b10;
		3'b100: idx = 2'b11;
	endcase
end
endmodule : test
```
Synthesizing this with `dc_shell` results in inferred latches (ignore for a moment the fact that DC doesn't error out when latches are inferred in an `always_comb` block) because only three out of eight possible mux selects are handled in the switch statement.
This can be fixed by adding a default case, or by adding a default assignment prior to the switch statement.  However, this infers unnecessary logic when we are guaranteed the external controllability constraint that `onehot` is, in fact, one-hot.

However, adding the `unique` keyword solves this problem optimally:
```systemverilog
	unique case (onehot)
		3'b001: idx = 2'b01;
		3'b010: idx = 2'b10;
		3'b100: idx = 2'b11;
	endcase
```
The `unique case` is a mechanism for the RTL designer to encode controllability constraints.  It encodes the designer's guarantee to the synthesis tool / simulator / model checker that for each evaluation of the switch control signal, exactly one case in the switch statement will match.  Similarly, `unique` can be used to qualify `if-else if*-else` statements.

Other SV2009 keywords `unique0` and `priority` also modify the semantics of control statements.

dalance:
Closes #937


taichi-ishitani:
To emit case statements with `cond_type` safely but `cond_type` is ignored, we need to emit the last case item as `default` item.
(refs: https://github.com/veryl-lang/veryl/issues/937#issuecomment-2347386406)

However, its seems that current emitter is not implemented like that.
https://github.com/veryl-lang/veryl/blob/fb6ce038123a41da80dbc81887738a150c33918b/crates/emitter/src/emitter.rs#L332

dalance:
Yes. This is a reason why this PR is draft.
I don't have enoght time to progress this PR for some time, so I sent this to prevent to forget it.

Issue #1029, opened 2024-10-22
Pre-defined bit vector type


taichi-ishitani:
8/16/32/64 bits bit vectors are commonly used so I think it's useful that these types are defined ahead of time as the `std` library.

```systemverilog
package std_pkg {
  type ul8  = logic<8>;
  type ul16 = logic<16>;
  type ul32 = logic<32>;
  type ul64 = logic<64>;

  type il8  = signed logic<8>;
  type il16 = signed logic<16>;
  type il32 = signed logic<32>;
  type il64 = signed logic<64>;

  type ub8  = bit<8>;
  type ub16 = bit<16>;
  type ub32 = bit<32>;
  type ub64 = bit<64>;

  type ib8  = signed bit<8>;
  type ib16 = signed bit<16>;
  type ib32 = signed bit<32>;
  type ib64 = signed bit<64>;
}
```

PR #1044, opened 2024-10-31
Add pre-defined vector types

taichi-ishitani:
close veryl-lang/veryl#1029


dalance:
Actual usage seems to be like below.
I think it is redundant, and `types` package may be sufficient.

```systemverilog
var a: $std::std_types_pkg::il32;

var a: $std::types::il32;
```

taichi-ishitani:
> Actual usage seems to be like below. I think it is redundant, and `types` package may be sufficient.
> 
> ```systemverilog
> var a: $std::std_types_pkg::il32;
> 
> var a: $std::types::il32;
> ```

I changed the package name.

Issue #1047, opened 2024-11-01
Generated filelist includes unused file


taichi-ishitani:
Generated filelist includes unused file like below.

```
/home/taichi/workspace/rggen/rggen-veryl-rtl/rggen_mux.sv
/home/taichi/workspace/rggen/rggen-veryl-rtl/rggen_rtl_pkg.sv
/home/taichi/workspace/rggen/rggen-veryl-rtl/rggen_bit_field_if.sv
/home/taichi/workspace/rggen/rggen-veryl-rtl/rggen_bit_field_w01trg.sv
/home/taichi/workspace/rggen/rggen-veryl-rtl/rggen_bit_field.sv
/home/taichi/workspace/rggen/rggen-veryl-rtl/rggen_register_if.sv
/home/taichi/workspace/rggen/rggen-veryl-rtl/rggen_address_decoder.sv
/home/taichi/workspace/rggen/rggen-veryl-rtl/rggen_register_common.sv
/home/taichi/workspace/rggen/rggen-veryl-rtl/rggen_indirect_register.sv
/home/taichi/workspace/rggen/rggen-veryl-rtl/rggen_default_register.sv
/home/taichi/workspace/rggen/rggen-veryl-rtl/rggen_bus_if.sv
/home/taichi/workspace/rggen/rggen-veryl-rtl/rggen_external_register.sv
/home/taichi/workspace/rggen/rggen-veryl-rtl/dependencies/std/coding/test_linear_sec.sv
```

`test_linear_sec.sv` is a unit test for std library so it should not be included.

PR #1064, opened 2024-11-08
Remove files including test only from generated filelist

taichi-ishitani:
fix veryl-lang/veryl#1047


dalance:
I added a test for filelist.
The following line is expected to remove by this PR.
https://github.com/veryl-lang/veryl/blob/0ad6ebd60be8798633d86abe42abfad28d30034e/crates/tests/src/lib.rs#L332

Issue #1122, opened 2024-12-04
No error is reported for an anonymous identifier used as a factor


taichi-ishitani:
Veryl compiler reports no error for an anonymous identifier used as a factor and emits invalid SV code.

```systemverilog
module ModA {
  var a: logic;
  assign a = _;
}
```

![image](https://github.com/user-attachments/assets/42382205-cc5f-455e-b1f6-cdd8164e173e)

PR #1124, opened 2024-12-05
Restrict use of anonymous identifier

taichi-ishitani:
Currently, anonymous identifier (`_`) can be used at any location.
This is invalid situation and caseses #1122.

This PR is to add limitation to use of anonymous identifier to fix this situation.
Anonymous identifier can be used only for:
- clock domain
- unconnected output port

fix veryl-lang/veryl#1122


crates/analyzer/src/tests.rs
```diff
@@ -1656,6 +1656,52 @@ fn undefined_identifier() {
...
+    let code = r#"
+    module ModuleA {
+        let a: logic = _ + 1;
+    }
+    "#;
+
+    let errors = analyze(code);
+    assert!(matches!(
+        errors[0],
+        AnalyzerError::UndefinedIdentifier { .. }
```
dalance, line 1669: I think `AnonymousIdentifierUsage` is better than `UndefinedIdentifier` for this usage.
This is because people who read "`_` is undefined" will think "I must define `_`".

taichi-ishitani: I changed the error class `UndefinedIdentifier` to `AnonymousIdentifierUsage`.

Issue #1131, opened 2024-12-07
generic interface file is empty 


nananapo:
If there is a function, module/interface name is disappeared.

https://github.com/nananapo/vb/tree/main/1131

mmio_controller.veryl
```mmio_controller.veryl
import eei::*;

module mmio_controller::<DATA_WIDTH: const, ADDR_WIDTH: const> (
    clk: input clock,
    rst: input reset,
) {
    inst mb: membus_if::<DATA_WIDTH, ADDR_WIDTH>;
    function nop () {}
}
```
mmio_controller.sv
```mmio_controller.sv
module core___mmio_controller__eei_MEM_DATA_WIDTH__eei_MEM_ADDR_WIDTH
    import core_eei::*;
 (
    input logic clk,
    input logic rst
);
     mb (); // <- ??????
    function automatic void nop() ;
    endfunction
endmodule
```
membus_if.sv
```membus_if.sv
// EMPTY!!!!
```

...and filelist order is incorrect

```core.f
/home/kanataso/Documents/vb/1131/src/membus_if.sv <- 
/home/kanataso/Documents/vb/1131/src/eei.sv <-
/home/kanataso/Documents/vb/1131/src/mmio_controller.sv
/home/kanataso/Documents/vb/1131/src/top.sv
```

taichi-ishitani:
I open PR #1138 to fix invalid instance declaration and empty interface definition but the fix for wrong filelist order is not included.
I think this is the same issue with #1070.

PR #1139, opened 2024-12-11
Fix wrong filelist order

### LINKED ISSUE #1070: Reference through expression is not affected to filelist order
* package.veryl

```systemverilog
package PackageA {
    const A: u32 = 1;
}
```

* module.veryl

```systemverilog
module ModuleA {
    let a: u32 = PackageA::A;
}
```

The filelist order should be `package.veryl, module.veryl`, but it is not actually.

taichi-ishitani:
This PR is to fix #1070.

Currently, type dag does not include package and enum referred in expression.
This causes #1070 because filelist is sorted based on type dag.
Therefore, we need to put such package and enum to type dag to get correct filelist order.


taichi-ishitani:
filelist showing #1131 has not been resolved yet.
Need more investigation.

taichi-ishitani:
To resolve #1131, we need to resolve all generic argumetns before creating type dag.
However, resolving generic arguments and creating type dag are executed in parallel.

taichi-ishitani:
#1144 also includes fix for #1070 so this will be closed.

PR #1154, opened 2024-12-18
Refactor `msb` check and emittion

### LINKED ISSUE #1142: resolving msb of struct member is failed
```sv
module TopModule {
    struct A {
        a: logic<6>,
    }
    let _a: A        = 0;
    let _b: logic<5> = _a.a[msb:1];
}
```

```
]8;;https://doc.veryl-lang.org/book/07_appendix/02_semantic_error.html#unknown_msb\unknown_msb (link)]8;;\

  💥 resolving msb is failed
   ╭─[:6:29]
 5 │     let _a: A        = 0;
 6 │     let _b: logic<5> = _a.a[msb:1];
   ·                             ─┬─
   ·                              ╰── Error location
 7 │ }
   ╰────
  help: 
```

taichi-ishitani:
This PR is to fix following two issues of `msb` feature.

* `msb` cannot be used on struct/union member, enum and paramter
* Wrong msb position calcuration for a type defined in outside of the current scope
  * https://github.com/veryl-lang/veryl/issues/1142#issuecomment-2547700428
  * use `$size` and `$bits` system function instead of evaluating the size declaration

fix #1142


nananapo:
Why is it expected to cause UnknownMSB when using ```msb``` on an interface variable?
It can also be resolved by using $size.
https://github.com/veryl-lang/veryl/blob/6b2a69650d26fbdd8757807eabbbd05693960922/crates/analyzer/src/tests.rs#L3698-L3729

Issue #1126, opened 2024-12-07
Self-reference in always_comb causes many warnings in other files


clkbug:
veryl version: 03146a08

```SystemVerilog
module a (
    x: output logic<32>,
) {
    always_comb {
        x = x; // this causes many warnings
    }
}
```

The followings are the first three warnings (actually 24 warnings):

```
Error:   × veryl check failed

Warning: unassign_variable (link)

  ⚠ x is unassigned
   ╭─[/tmp/bug/a.veryl:2:5]
 1 │ module a (
 2 │     x: output logic<32>,
   ·     ┬
   ·     ╰── Error location
 3 │ ) {
   ╰────
  help: 
Warning: unassign_variable (link)

  ⚠ x is unassigned
   ╭─[/tmp/bug/a.veryl:1:16]
 1 │ import async_fifo_pkg::calc_default_depth;
   ·                ┬
   ·                ╰── Error location
 2 │ 
   ╰────
  help: 
Warning: unassign_variable (link)

  ⚠ x is unassigned
   ╭─[/tmp/bug/a.veryl:1:16]
 1 │ package async_fifo_pkg {
   ·                ┬
   ·                ╰── Error location
 2 │     function calc_default_depth (
   ╰────
  help: 
Warning: unassign_variable (link)

  ⚠ x is unassigned
   ╭─[/tmp/bug/a.veryl:1:16]
 1 │ module async_fifo_reset_sync::<S: synchronizer> #(
   ·                ┬
   ·                ╰── Error location
 2 │     param MERGE_RESET      : bit = '0,
   ╰────
  help: 

```

taichi-ishitani:
x is not assigned so this is expected error.

PR #1157, opened 2024-12-19
Fix issue that `unassign_variable` error is reported many times unexpectedly

taichi-ishitani:
fix veryl-lang/veryl#1126

### DIFF
--- a/crates/analyzer/src/analyzer.rs
+++ b/crates/analyzer/src/analyzer.rs
@@ -11,10 +11,11 @@ use crate::symbol::{
 use crate::symbol_table;
 use crate::type_dag;
 use crate::var_ref::{
-    AssignPosition, AssignPositionTree, AssignPositionType, ExpressionTargetType,
+    AssignPosition, AssignPositionTree, AssignPositionType, ExpressionTargetType, VarRef,
     VarRefAffiliation, VarRefPath, VarRefType,
 };
 use itertools::Itertools;
+use std::collections::HashMap;
 use std::path::Path;
 use veryl_metadata::{Build, Lint, Metadata};
 use veryl_parser::resource_table;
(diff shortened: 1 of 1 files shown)


crates/analyzer/src/analyzer.rs
```diff
@@ -168,26 +172,29 @@ impl<'a> AnalyzerPass3<'a> {
...
-                        }) {
-                            let full_path = var_ref.path.full_path();
-                            let symbol = symbol_table::get(*full_path.first().unwrap()).unwrap();
-                            ret.push(AnalyzerError::unassign_variable(
-                                &var_ref.path.to_string(),
-                                self.text,
-                                &symbol.token.into(),
-                            ));
-                        }
+        for (_, list) in self.var_refs.iter().filter(|(key, _)| {
```
dalance, line 175: This `for` seems to be too long. How about using tempolary value like below?

```rust
let var_refs = self.var_refs.iter().filter(...);
for (_, list) in var_refs {
}
```
```

File: crates/analyzer/src/analyzer.rs [hunk 2]
```diff
@@ -168,26 +172,29 @@ impl<'a> AnalyzerPass3<'a> {
...
+            matches!(
+                key,
+                VarRefAffiliation::AlwaysComb { token } if token.source == self.path
+            )
+        }) {
+            let list: Vec<_> = list.iter().enumerate().collect();
+            let assign_list: Vec<_> = list.iter().filter(|(_i, x)| x.is_assign()).collect();
+            for (i, var_ref) in &list {
+                if let VarRefType::ExpressionTarget { r#type } = var_ref.r#type {
+                    if matches!(
```
dalance, line 185: Could you give an appropriate name to the `if` condition?

taichi-ishitani:
I fixed them.

Issue #963, opened 2024-09-13
Interface prototype


dalance:

dalance:
Interface prototype may be the same as concrete interface in almost cases. 

```systemverilog
proto interface ProtoA #(
    param X: u32,
) {
    var a: logic;

    modport A {
        a: input,
    }
}

interface InterfaceA for ProtoA #(
    param X: u32,
) {
    var a: logic;

    modport A {
        a: input,
    }
}
```

So generic bound by interface may be sufficient instead of adding interface prototype.

```systemverilog
block rggen_connect_bit_field_if::<RIF: InterfaceA, FIF: InterfaceA, LSB: const, WIDTH: const> {
    assign  FIF.valid                 = RIF.valid;
    assign  FIF.read_mask             = RIF.read_mask[LSB+:WIDTH];
    assign  FIF.write_mask            = RIF.write_mask[LSB+:WIDTH];
    assign  FIF.write_data            = RIF.write_data[LSB+:WIDTH];
    assign  RIF.read_data[LSB+:WIDTH] = FIF.read_data;
    assign  RIF.value[LSB+:WIDTH]     = FIF.value;
}
```

taichi-ishitani:
The `prototype` generics boundary is to specify a module name which will be instantiated.
This notation may be confusable because its appearance is similar with the `prototype` boundary but thier functionality are different.
Therefore, I think other notation is needed like below.

```
block rggen_connect_bit_field_if::<RIF: inst InterfaceA, FIF: inst InterfaceA, LSB: const, WIDTH: const> {
    assign  FIF.valid                 = RIF.valid;
    assign  FIF.read_mask             = RIF.read_mask[LSB+:WIDTH];
    assign  FIF.write_mask            = RIF.write_mask[LSB+:WIDTH];
    assign  FIF.write_data            = RIF.write_data[LSB+:WIDTH];
    assign  RIF.read_data[LSB+:WIDTH] = FIF.read_data;
    assign  RIF.value[LSB+:WIDTH]     = FIF.value;
}
```

taichi-ishitani:
interface can have functions so I think interface prototype is meaningful.

```systemverilog
proto interface InterfaceA {
  function FuncA(a: input logic, b: input logic) -> logic;

  modport mp {
    FuncA: import,
  }
}
```

PR #1192, opened 2025-01-08
Introduce `inst` generic boundary

taichi-ishitani:
This PR is to a new generic boundary named `inst`.
The `isnt` boundary is to specify module or interface instance as a generic parameter.

refs: https://github.com/veryl-lang/veryl/issues/963#issuecomment-2574405689


crates/analyzer/src/handlers/check_type.rs
```diff
@@ -210,12 +210,62 @@ impl VerylGrammarTrait for CheckType<'_> {
                                         ));
                                     }
                                 }
+                                GenericBoundKind::Inst(proto) => {
+                                    let proto_match = if arg.is_resolvable() {
```
dalance, line 214: At Rust 2024 (Rust 1.85), if-let chain will be stabilized.
After it, this large if chain can be re-written like below. So `TODO` comment may be useful to mark for re-writing.

```rust
if arg.is_resolvable()
  && let Ok(symbol) = ...
  && let SymbolKind::Instance(x) = ...
  && let Ok(x) = ...
  && let Some(ref x) = x.found.proto() {
  ...
} else {
  false
}
```

testcases/veryl/55_generic_module.veryl
```diff
@@ -41,3 +42,9 @@ module Module55H::<W: const> {
 
     let _a: StructH::<W> = 0;
 }
+
```
dalance, line 45: I think this example is bit strange.
`IF` is bound by `Interface55I`, but `Interface55I` is not `proto` but specific type, so `IF` is always `Interface55I`.
Therefore I think generics is not required in this case.
How about adding `proto interface` support described at #963 like below?

```systemverilog
proto interface PInterface {
    function FuncA(a: input logic, b: input logic) -> logic;
}

interface Interface55I for PInterface {
    function FuncA(a: input logic, b: input logic) -> logic {
        return 0;
    }
}

module Module55I::<IF: PInterface> {
    inst u: IF;
}

module X {
    inst u: Module55I::<Interface55I>;
}
```

taichi-ishitani: I agree with you but a lot of changes is needed to introduce `proto interface`.
For now, how about removing this example and opening a new PR to introduce `proto interface`?

dalance: I agree you. `check_type.rs:387-390` and `symbol.rs:304` should be removed probably.

taichi-ishitani: I removed changes to allow interface to be used as generic boundary.

crates/analyzer/src/handlers/check_type.rs
taichi-ishitani: I added todo comments.

https://github.com/veryl-lang/veryl/blob/c353e6138556dd4d5c13571e37eb9b3cbce7700d/crates/analyzer/src/handlers/check_type.rs#L214
https://github.com/veryl-lang/veryl/blob/c353e6138556dd4d5c13571e37eb9b3cbce7700d/crates/analyzer/src/handlers/check_type.rs#L267

crates/analyzer/src/symbol.rs
```diff
@@ -297,6 +297,18 @@ impl Symbol {
             _ => Vec::new(),
         }
     }
+
+    pub fn proto(&self) -> Option<SymbolPath> {
+        match &self.kind {
+            SymbolKind::Module(x) => x.proto.clone(),
+            SymbolKind::Interface(_) => Some((&self.token).into()),
```
dalance, line 304: Is this line necessary?

taichi-ishitani: Yes.
It is used at here.
https://github.com/veryl-lang/veryl/blob/c353e6138556dd4d5c13571e37eb9b3cbce7700d/crates/analyzer/src/handlers/check_type.rs#L274

dalance: In this PR, interface can't be used as generic boudary. So I think adding ` SymbolKind::Interface(_)` is not necessary, and `SymbolKind::Interface(x) => x.proto.clone(),` should be added at introducing proto interface.

taichi-ishitani: The `inst` boundary with specific interface is needed for usecase shown in #793.
Therefore, this change is needed.

dalance: I understood this line is used, but I think it is not appropriate that non proto symbol returns `Some` from `proto()`.
Changing `proto()` to `proto_or_inst()` or splitting it into `proto()` and `inst()` is better.
I prefer the splitting way because inst can be assigned to a generic parameter bound as proto if proto and inst are not distincted.

taichi-ishitani: I removed the statement for the `SymbolKind::Interface` but did not add the `inst` function.
Instead of the function, I added `resolve_inst_generic_arg_type` and `resolve_proto_generic_arg_type` functions to simplify resolving type of generic arguments.

dalance: Thanks!
I think this is better than my idea.

PR #1199, opened 2025-01-10
Allow benches on non-Linuxes

jsinger67:
Since crate pprof is only available on linux the module perf only compiles on linux.
This prevents executing benches on other OSes.
Can confirm that this works on Windows.


dalance:
Thanks!
I'll merge this PR.

PR #1226, opened 2025-02-05
[Fix] Basic Support for File Renaming and Deleting

### LINKED ISSUE #1224: [Bug] Module is marked as duplicated after moving the source file
Hello,   
I'm currently using Veryl Language Server with Veryly VSCode extension.  

It occurred that after moving a source file from a directory to another, the module and all its members (the `logic`s and `const`s) are marked as duplicated.  `Semantic Error: ModuleA is duplicated` as an instance.  

These errors are gone after executing `vscode-veryl: Restart Language Server`.

I'm working on VSCode with Veryl extension version `0.13.5`.  
My system is Windows 11 24H2.

Hamster5295:
Closes #1224

### DIFF
--- a/crates/languageserver/src/server.rs
+++ b/crates/languageserver/src/server.rs
@@ -5,7 +5,6 @@ use futures::executor::block_on;
 use ropey::Rope;
 use std::collections::VecDeque;
 use std::path::{Path, PathBuf};
-use std::str::FromStr;
 use tower_lsp::lsp_types::*;
 use tower_lsp::Client;
 use veryl_analyzer::namespace::Namespace;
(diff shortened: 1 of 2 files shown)


crates/languageserver/src/server.rs
```diff
@@ -31,6 +32,16 @@ pub enum MsgToServer {
         version: i32,
     },
     DidChangeConfiguration(ServerConfigItem),
+    WillRenameFile {
+        old_url: String,
```
dalance, line 36: I can understand the reason that type of url is `String` in newly added variants, it is came from `lsp_types::FileRename`.
But `Url` may be more appropriate, because it is more consistent with the name of field, and finally it is converted to `Url` in `did_rename_files`.

dalance:
I think logic of the code has no problem, and commented a point about type definition.

crates/languageserver/src/server.rs
Hamster5295: Get it!

dalance:
Looks good!
I'll merge this PR.

Issue #1229, opened 2025-02-06
Miss-formatted SV when omiting port connections


taichi-ishitani:
Format of generated SV is incorrect when port connections are omitted.

Veryl code

```systemverilog
      inst u_bit_field: rggen::rggen_bit_field #(
        WIDTH:          4,
        INITIAL_VALUE:  INITIAL_VALUE,
        SW_WRITE_ONCE:  0,
        TRIGGER:        0
      )(
        i_clk:            i_clk,
        i_rst:            i_rst,
        bit_field_if:     bit_field_sub_if,
        o_write_trigger:  _,
        o_read_trigger:   _,
        o_value:          o_register_0_bit_field_0
      );
```

Generated SV

```systemverilog
            rggen_rggen_bit_field #(
                .WIDTH         (4            ),
                .INITIAL_VALUE (INITIAL_VALUE),
                .SW_WRITE_ONCE (0            ),
                .TRIGGER       (0            )
            ) u_bit_field (
                .i_clk           (i_clk                   ),
                .i_rst_n         (i_rst_n                 ),
                .bit_field_if    (bit_field_sub_if        ),
                .o_write_trigger (                        ),
                .o_read_trigger  (                        ),
                .o_value         (o_register_0_bit_field_0),
                .i_sw_write_enable ('1),
                .i_hw_write_enable ('0),
                .i_hw_write_data   ('0),
                .i_hw_set          ('0),
                .i_hw_clear        ('0),
                .i_value           ('0),
                .i_mask            ('0),
                .
                o_value_unmasked ()

            );
```

PR #1237, opened 2025-02-10
Fix miss fomratted SV

taichi-ishitani:
Emission of omitted ports is copy of port declarations. Therefore, location info of inst declaration and such port connections are different and this causes such miss emitted SV.
To fix it, location info needs to be adjusted.

fix veryl-lang/veryl#1229


taichi-ishitani:
Expression portions are not aligned because of the same reason but I'm not sure how to adjust it.
https://github.com/veryl-lang/veryl/blob/5e607c5a9986e70326a8a9d252a669f8eed3bf42/testcases/sv/73_port_default_value.sv#L29-L49

dalance:
I think that changing aligner structure may be necessary.
Aligner ties the amount of align with the token location, but default port value doesn't have the corresponding token.
I'll consider new aligner architecture.

dalance:
I'm working on #1245.
In the case of default value only, it was fixed, but combination of explicit and default is not yet.

dalance:
I'll merge this PR at first, and the remaining issue will be fixed by merging #1245.

PR #1388, opened 2025-03-12
Add new waveform target / Implement waveform for cocotb

hellow554:
This is a 2 in 1 (please tell me if I should split this up) merge.

This changes the option of `veryl test --wave`. You now have to provide either `vcd` or `fst` as a target format.

Also cocotb now supports waveform generation since it's the (my) main target of testing.

### DIFF
--- a/crates/metadata/src/test.rs
+++ b/crates/metadata/src/test.rs
@@ -14,6 +14,8 @@ pub struct Test {
     pub vivado: VivadoProperty,
     #[serde(default)]
     pub waveform_target: WaveFormTarget,
+    #[serde(default)]
+    pub waveform_format: WaveFormFormat,
 }
 
 #[derive(Clone, Copy, Debug, Default, Serialize, Deserialize, PartialEq, Eq)]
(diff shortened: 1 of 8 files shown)


crates/metadata/src/test.rs
```diff
@@ -65,3 +65,22 @@ pub enum WaveFormTarget {
     #[serde(rename = "directory")]
     Directory { path: PathBuf },
 }
+
+#[derive(Clone, Copy, Default, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(tag = "type")]
+pub enum WaveFormFormat {
```
dalance, line 71: How about adding `waveform_format: WaveFormFormat` to `Test`, and supporting `veryl test --wave` which uses default waveform format defined in `Veryl.toml`?
So user can use simply `veryl test --wave`, and can specify `veryl test --wave fst` as needs.

crates/veryl/src/runner/cocotb.rs
```diff
@@ -163,6 +163,16 @@ impl Runner for Cocotb {
 
         let module = format!("{}_{}", metadata.project.name, top.unwrap());
 
+        let (py_waves, args) = if let Some(WaveFormFormat::Vcd) = wave {
```
dalance, line 166: For adding another waveform format in future, using `match` instead of `if let` may be better.

dalance (commented):
I agree this PR overall ( 2 in 1 is no problem ).

I added some comments. Could you check it?

crates/metadata/src/test.rs
hellow554: I introcuded the new field in `Veryl.toml` but discarded my initial idea of specifying the target format via cli.
Is that ok for you?

hellow554: I added the field in the `Veryl.toml` but discarded my initial idea of specifiying it via cli.

Is that ok for you?

crates/veryl/src/runner/cocotb.rs
```diff
@@ -176,11 +186,14 @@ runner.build(
     verilog_sources=sources,
```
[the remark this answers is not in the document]
dalance: This is because I checked verilator only.
If you use another simulator, I'll accept a PR.

dalance:

PR #1410, opened 2025-03-18
veryl-std: Implementation of AXI3 / AXI4 interface

janschiefer:
This is a implementation for a System Verilog interface of an AXI3 and AXI4 interface.

Modports: master, slave, monitor, master_interconnect and slave_interconnect (the last two with bus clock and reset output).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4852,9 +4852,9 @@ checksum = "1f3ccbac311fea05f86f61904b462b55fb3df8837a366dfc601a0161d0532f20"
 
 [[package]]
 name = "tokio"
-version = "1.44.1"
+version = "1.44.2"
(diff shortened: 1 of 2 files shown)


taichi-ishitani:
AXI protocol has a lot of configurable parameters and I think they should be managed together.
How about declaring these parametes in a pacakge and passing the package to the interface as generic argument?
Such package can also manage tyepdefs defined from the parameters together.

janschiefer:
Sure. I'll update the commit to your specifications in the next week's.

dalance:
@taichi-ishitani  @janschiefer 

I wrote some issues seeing your PR. #1414 #1415 
Could you tell me your opinion about them?

Issue #1399, opened 2025-03-14
Introduce `connect` operation


taichi-ishitani:
refs #1391

The `connect` operator `<>` is to connect two interfaces.

```systemverilog
always_comb {
  master_if <> slave_if;
}
```

The above code is expanded into each port connections like below.

```systemverilog
always_comb begin
  slave_if.ready  = master_if.ready;
  master_if.valid = slave_if.valid;
  master_if.data  = slave_if.data;
end
```

Modport is required to determine directions if an operand is an interface instance.

```systemverilog
always_comb {
 // if_a is an interface instance
 // slave_if is a modport
  if_a.master <> slave_if;
}
```

`connect` keyword is required if a `connect` operation is used in a module body but not an `always_*` procedure.

```systemverilog
module ModuleA {
  connect if_a.master <> slave_if;
}
```

For this case, modport including `intou` ports and an arrray interface instance/modport can be used as an operand.
Such operations will be emitted like below.

* Including inout ports

```systemverilog
tran (if_a.scl, if_b.scl);
tran (if_a.sda, if_b.sda);
```

* Array interface/modport

```systemverilog
for (genvar __i = 0;__i < 4;__i++) begin : __g_connect_if_a_if_b
  alwasy_comb begin
    slave_if[__i].ready = master_if[__i].ready;
    master_if[__i].valid = slave_if[__i].valid;
    master_if[__i].data  = slave_if[__i].data;
  end
end
```

There are some needs to tie output ports of an interface to a specific value.
(e.g. https://github.com/pezy-computing/pzbcm/blob/985e7147ab42682e62f32604433090b3be5f8482/pzcorebus_dummy/pzcorebus_dummy_slave.sv#L36-L47)
`connect` operation should also support such cases.

```systemverilog
always_comb {
  slave_if <> 0;
}
```

This code will emitted like below.

```systemverilog
always_comb begin
  slave_if.ready = 0;
end
```

taichi-ishitani:
To connect array interfaces, array sizes of these interfaces need to be matched.
For this purpose, expression comparator is required.
The comparator compares two expression have the same operands and operator does not evaluate expressions.

taichi-ishitani:
For array interface connection, part connection should be supported.

```systemverilog
connect a_if[0+:4].master <> slave_if[4+:4];
```

dalance:
[...]

Issue #1391, opened 2025-03-12
Autoconnect for interfaces?


janschiefer:
Since Veryl focuses on simplicity and a good developer experience, mechanisms to reduce (error prone) boilerplate code seen desirable.

Why not implement a operator that automatically connects complementary (input - output, output- input, inout-inout?) elements between same type interfaces? 

```
interface axi_if {

var AWREADY;
var AWVALID;
// imagine 20 more fields here...

modport interconn_slave {
AWREADY : output,
AWVALID : input,
//...
}

modport interconn_master {
..converse(interconn_slave)
}

}

module interconn(
bus_in: modport axi_if::interconn_slave,
bus_out: modport axi_if::interconn_master
)
{

assign bus_in.AWVALID = bus_out.AWVALID;
assign bus_out.AWREADY = bus_in.AWREADY;
//Imagine 20 more assignments here...

//this simply becomes:
bus_in <> bus_out;

}

```

(And yes, I also like SpinalHDL, but the Verilog code generated is horrible.)

taichi-ishitani:
Related issue
https://github.com/veryl-lang/veryl/issues/961

taichi-ishitani:
> Related issue [#961](https://github.com/veryl-lang/veryl/issues/961)

How about the idea below?

Define a function which can take a interface modport as a parameter and connect two interfaces by using this function.

```systemverilog
interface axi_if {
  var awready: logic;
  var awvalid: logic;

  function connect (
    self,
    slave_if: axi_if::slave,
  ) {
    slave_if.awready = self.awready;
    self.awvalid     = slave_if.awvalid;
    ...
  }

  modport master {
    awready: input,
    awvalid: output,
    connect: import,
  }
}

inst master_if: axi_if;
inst slave_if:  axi_if;

assign master_if.connect(slave_if);
```

SystemVerilog does not support functions which take interface instances or modports.
Therefore, we need to emit body code of such function into SystemVerilog directoly.

Emitted SV

```systemverilog
axi_if master_if();
axi_if slave_if();

always_comb begin
  slave_if.awready = master_if.awready;
  master_if.awvalid = slave_if.awvalid;
  ...
end
```

dalance:
I think funciton associated interface is useful, but special syntax for connection may be better because it occurs frequently.
Probably inferring direction by associated function is difficult for the following case.

```systemverilog
module ModuleA (
    port: modport Interface::slave,
) {
    inst temp0: Interface;
[...]

PR #1423, opened 2025-03-22
Introduce `connect` operation

taichi-ishitani:
close veryl-lang/veryl#1399
close veryl-lang/veryl#1391

----

In the current implementation, the following conditions are used to check which ports are available to connect:

* match port names
* directions are connectable
    * input/output
    * inout/inout

Implementation is here.
https://github.com/veryl-lang/veryl/blob/966c01d028b9429f25b88adb9c9d9ef9acd4740c/crates/analyzer/src/connect_operation_table.rs#L85-L102

----

Due to https://github.com/veryl-lang/veryl/issues/1399#issuecomment-2731350239, only non array modport and selected single element of array modport can be used as an operand.


crates/analyzer/src/symbol.rs
```diff
@@ -965,18 +973,46 @@ impl TypeKind {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone)]
 pub struct UserDefinedType {
     pub path: Vec<StrId>,
+    pub identifier: syntax_tree::ScopedIdentifier,
```
dalance, line 979: How about changing `path: Vec<StrId>` to `path: GenericSymbolPath` instead of adding `identifier`?
`GenericSymbolPath` probably have sufficient information, and is more structured than `ScopedIdentifier`.

crates/analyzer/src/handlers/check_expression.rs
```diff
@@ -307,9 +308,31 @@ impl VerylGrammarTrait for CheckExpression {
...
+                        } else if matches!(
+                            connect_operation.unwrap().rhs,
+                            ConnectOperand::Expression(_)
+                        ) {
+                            (true, false)
+                        } else {
+                            (false, false)
+                        };
+
+                        if !do_check.0 {
```
dalance, line 325: Could add comment for the meaning of `do_check.0` and `do_check.1` or change them to struct?

taichi-ishitani:
Thank you for your feedbacks. I fixed them.

PR #1453, opened 2025-03-28
Add elsif/els attribute

### LINKED ISSUE #1350: `elsif` and `els` attribute
#1349 will add the way to judge define's exclusivity.
So the way to express exclusivity becomes important.

For example, the following code is multi-drive error, because `FEATURE_A`, `FEATURE_B` and `FEATURE_C` are not exclusive.

```systemverilog
#[ifdef(FEATURE_A)]
assign a = 1;
#[ifdef(FEATURE_B)]
assign a = 2;
#[ifdef(FEATURE_C)]
assign a = 3;
```

The following code can express exclusivity.
In this case, each `assign` 's define context is exclusive, and not multi-drive error.
But this is too complicated.

```systemverilog
#[ifdef(FEATURE_A)]
assign a = 1;
#[ifndef(FEATURE_A)]
#[ifdef(FEATURE_B)]
assign a = 2;
#[ifndef(FEATURE_A)]
#[ifndef(FEATURE_B)]
#[ifdef(FEATURE_C)]
assign a = 3;
```

So `elsif` and `else` is useful like below:

```systemverilog
#[ifdef(FEATURE_A)]
assign a = 1;
#[elsif(FEATURE_B)]
assign a = 2;
#[elsif(FEATURE_C)]
assign a = 3;
```

`elsif` and `else` have some ambiguity.

```systemverilog
#[ifdef(FEATURE_A)]
assign a = 1;

let x: logic = 1; // this statement is not in both `ifdef` and `elsif`.

#[elsif(FEATURE_B)]
assign a = 2;
```

Such case should be treated as error.

dalance:
Closes #1350


dalance:
I changed the title of PR. The actucal change from `els` to `else` is #1495 .

Issue #1445, opened 2025-03-27
[Feature] Simlify `if` expression


taichi-ishitani:
The if expression is used as alternative for System Verilog's conditional operator so more simlified notation is needed.
For example:

```systemverilog
if a ? b : c

if a ? b : if c ? d : e

if a ? if b ? c : d : e
```

taichi-ishitani:
@dalance ,

Should the original notation be left?

dalance:
The original notation is always redundant because statement can't be placed in `{}`.
So I think changing to new syntax is better.

```systemverilog
if a { x = y + 1; x + y } else { c }  // This can't be used
```

taichi-ishitani:
How notate `else if` branches? Which one is preferred?

```systemverilog
if a ? b :
if c ? d : e;
```

```systemverilog
     if a ? b
else if c ? d : e;
```

dalance:
I think the first is better because I think it is diffucult to see that `else` becomes `:` but `else if` is kept.

PR #1457, opened 2025-03-29
Simplify if expression notation

taichi-ishitani:
close veryl-lang/veryl#1445

Priority of `if` expression is lowest priority because priority of SV's conditional operation is so.
https://github.com/veryl-lang/veryl/blob/14a735df591d8332de7ebcc606ae54a09852dd3c/crates/parser/veryl.par#L443-L457


taichi-ishitani:
Time to generate the parser is increased due to definiton of `IfExpression`.
https://github.com/veryl-lang/veryl/blob/14a735df591d8332de7ebcc606ae54a09852dd3c/crates/parser/veryl.par#L443

I confirm that there are not impact in case of `Expression 01` is used instead of `Expression` like below.

```
IfExpression: { If Expression01 Question Expression01 Colon } Expression01;
```

Whici definition is preferred?

dalance:
> Priority of `if` expression is lowest priority because priority of SV's conditional operation is so.

I think the reason which SV's conditional operator is listed on precedence list is the following ambiguity.

```
a + b ? c : d  -> (a + b) ? c : d   or a + ( b ? c : d )
```

But if expression has not such ambiguity because it starts with `if` keyword.

```
a + if b ? c : d  -> a + ( b ? c : d )
```

So I think `IfExpression` can be placed in `Factor` as the same as the current implementation.

taichi-ishitani:
Parser generation did not complete after more than 10 minutes when the new if expression is added as a factor by the rule below.

```
IfExpression: If Expression Question Expression { Colon If Expression Question Expression } Colon Expression;
```

Therefore, I inserted the new if expression into the expression chain.

dalance:
How about the following?

```
IfExpression: If Expression Question Expression Colon Expression;
```

taichi-ishitani:
> How about the following?
> 
> ```
> IfExpression: If Expression Question Expression Colon Expression;
> ```

Thanks! I will try it.

taichi-ishitani:
> > How about the following?
> > ```
> > IfExpression: If Expression Question Expression Colon Expression;
> > ```
> 
> Thanks! I will try it.

Unfortunity, parser generation didn't complete with this rule.

dalance:
I understood it. The end of `Expression` can't be determined.
So the following is OK, for example.

```
IfExpression: If Expression Question Expression Colon Expression Colon;
```

I thought some case with terminating by a symbol.
But chaining of if expression is hard to see.

```
if a ? b : c:
if a ? b : if c ? d : e : : 
if (a ? b : c)
if (a ? b : if ( c ? d : e ) )
```

So I think your first implementation is better.
About the second definition, I wonder whether the above definition prohibit the following code.

```
if a ? if b ? c : d : e
```

taichi-ishitani:
> About the second definition, I wonder whether the above definition prohibit the following code.

Yes. This example causes parse error.
I feel difficult to see the end of the inner if expression and the second rule can force to wrap the inner if expression with `()` like below.

```
if a ? (if b ? c : d) : e
```

dalance:
I feel it is strange that `if a ? if b ? c : d : e` becomes syntax error.
It should be syntactically valid, but it should be fixed from the point of view of readability.
Users who don't read the definition of `veryl.par` can't understand why it is syntax error.

taichi-ishitani:
> Users who don't read the definition of veryl.par can't understand why it is syntax error.

Ok, I understand.
Inner if expressions are hard to read so I think the linter should report warnings for them.

dalance:
> Inner if expressions are hard to read so I think the linter should report warnings for them.

Yes. I agree it.

taichi-ishitani:
> Inner if expressions are hard to read so I think the linter should report warnings for them.

I implemented the checker.
https://github.com/veryl-lang/veryl/compare/51182ec03b1228da61c2fba661ff9e753ed0c052..1256e2459e45f59e98d04bda51a6f645610c15b2

Issue #1459, opened 2025-03-30
[Feature] Proto alias declaration


taichi-ishitani:
There are some cases that module which will be instantiated and related parameters want to be managed together.
(E.G. SRAM cell and its address width/data width paramters)

Introduce `proto alias declaration` into proto package declaration.
By introducing it, this kind of usecases can be supported like below.

```systemverilog
proto package ProtoPkg {
  alias module SRAM_CELL: sram_cell_wrapper_proto;

  const ADDRESS_WIDTH: u32;
  const DATA_WIDTH   : u32;
}

module sram_fifo_cell_wrapper::<PKG: ProtoPkg> (
  i_write_address: input  logic<PKG::ADDRESS_WIDTH>,
  i_read_address : input  logic<PKG::ADDRESS_WIDTH>,
  i_data         : input  logic<PKG::DATA_WIDTH>   ,
  o_data         : output logic<PKG::DATA_WIDTH>   ,
){
  inst u_sram_cell: PKG::SRAM_CELL (
    i_write_address,
    i_read_address ,
    i_data         ,
    o_data         ,
  );
}
```

PR #1473, opened 2025-04-01
Introduce proto alias declaration

taichi-ishitani:
close veryl-lang/veryl#1459


taichi-ishitani:
https://github.com/veryl-lang/veryl/actions/runs/14218367731/job/39840188255

I think this regression failure is caused by failure of creating or removing a lock file but I don't know how to fix it.

Issue #1484, opened 2025-04-03
Need struct/union/enum defined in interface?


taichi-ishitani:
Struct/union/enum defined in interface is a one of big problem to implement connect operation and modport expansion because such types can't be exported from the interface.
I'd like to remove Struct/union/enum definitions from interface declarations if there are no usecases.

dalance:
I checked the existing Veryl projects in GitHub through https://github.com/veryl-lang/discovery, and confirmed no project uses it.
So I agree this change.

PR #1521, opened 2025-04-11
Unsupport enum/struct/union declarations in interface declaration

taichi-ishitani:
close veryl-lang/veryl#1484


dalance:
This change disables enum/struct/union in generate block of module too.
I don't have such use case, but I feel the restriction is not clear for users.
About enum in interface, semantic error like "Enum can't be used in interface" is better than only "Syntax Error".

So how about reverting syntax and adding analyzer check?

taichi-ishitani:
@dalance 

Thank you for your review. I fixed your feedback.

PR #1532, opened 2025-04-16
Introduce interface prototype

taichi-ishitani:
close veryl-lang/veryl#963


dalance:
LSP test detect a parse error, and the error location is affected by syntax definition.
So the change is valid.

Issue #1536, opened 2025-04-18
Support u16/i16/u8/i8 fixed types


taichi-ishitani:
Support u16/i16/u8/i8 fixed types.
They will be translated to SystemVerilogs's `shortint unsigned`/`shortint signed`/`byte unsigned`/`byte signed`.

PR #1537, opened 2025-04-18
Introduce u8, u16, i8 and i16 fixed types

taichi-ishitani:
close veryl-lang/veryl#1536


taichi-ishitani:
These types have alaready been documented.
https://github.com/veryl-lang/doc/pull/21

Issue #1533, opened 2025-04-16
[Feature] Typed `const` generic boundary


taichi-ishitani:
Sometime we may want to specify a type to a `const` generic boundary.
For example:
A `const` generic boundary specifying an operation mode and provided by an enum variant.

```systemverilog
enum access_mode {
  READ_WRITE,
  READ_ONLY,
  WRITE_ONLY,
}

module bit_field::<ACCESS_MODE: const> {

}

inst u_bit_field: bit_field::<access_mode::READ_WRITE>;
```

I think such generic parameters should be typed like below and the type of the given argument should be checked.

```systemverilog
module bit_field::<ACCESS_MODE: const access_mode> {
}
```

dalance:
The current `const` assumes anything like integral number implicitly.
So using more accurate type may be better like below.
If all parameters become typed, `const` keyword is not necessary because they are always constant.

```systemverilog
module bit_field::<N: u32> {
}

inst u_bit_field: bit_field::<64'hffffffffffffffff>; // overflow can be detected
```

PR #1548, opened 2025-04-22
Support typed proto generic boundary

taichi-ishitani:
close veryl-lang/veryl#1533


crates/parser/veryl.par
```diff
@@ -778,14 +778,16 @@ WithParameterItem: ( Param | Const ) Identifier Colon ( ArrayType | Type ) Equ E
 GenericBound: Const
```
dalance, line 778: Is there any reason to keep `Const`?

taichi-ishitani: It is just compatibility for existing code.

dalance: I think it is better to introduce breaking change now than later.

taichi-ishitani: OK, I understood.
I will rmeove `const` generic boundry in this PR.

taichi-ishitani: > I will rmeove const generic boundry in this PR.

I've removed `const` generic boundary.
ae4aa6a1257773eb3f91ea0ced63b71e3f5d4083

PR #1554, opened 2025-04-25
Allow to use modport as function argument

taichi-ishitani:
close veryl-lang/veryl#961


taichi-ishitani:
I will implement connect statement support in function after mmerging this PR.

taichi-ishitani:
This PR also implements connect statement in function.

Issue #1403, opened 2025-03-15
[Feature] Serialize multi demensional array interface/modport into single dimensional array


taichi-ishitani:
Some EDA tools do not support multi dimensional array interface instance/modport.
For these tools, an option to serialize such interface/modport into single array is useful.

Veryl

```systemverilog
inst if_a: InterfaceA[2, 3];
```

Generated SV

```systemverilog
InterfaceA if_a[6]();
```

PR #1566, opened 2025-04-30
Support flattened array modport/instance

taichi-ishitani:
close veryl-lang/veryl#1403

Multi-demensional array instance/modport is flattened into a single demensional array instance/modport when the `flatten_array_interface` build option is set.

### DIFF
--- a/crates/analyzer/src/symbol_path.rs
+++ b/crates/analyzer/src/symbol_path.rs
@@ -585,6 +585,19 @@ impl From<&syntax_tree::WithGenericArgumentItem> for GenericSymbolPath {
     }
 }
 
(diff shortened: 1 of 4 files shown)


dalance:
I think the emitter code seems to be good, but I feel `serialize` is bit strange.

"flatten" may be more general?
https://numpy.org/doc/2.2/reference/generated/numpy.ndarray.flatten.html

taichi-ishitani:
> I think the emitter code seems to be good, but I feel `serialize` is bit strange.
> 
> "flatten" may be more general? https://numpy.org/doc/2.2/reference/generated/numpy.ndarray.flatten.html

I changed the option name to `flatten_array_interface`.

Issue #1574, opened 2025-05-02
Mangled path includes `$` character


taichi-ishitani:
Mangled path for generic component defined in std lib includes `$` character. `$` should be removed.

Veryl

```systemverilog
module Module68A (
    a_if: modport $std::std_test_if::<$std::StdTestPkg::<8>>::mp,
) {}
```

Generated SV

```systemverilog
module veryl_testcase_Module68A (
    std___std_test_if__$std___StdTestPkg__8.mp a_if
);
endmodule
```

PR #1576, opened 2025-05-02
Remove `$` character from project prefix of `std` lib

taichi-ishitani:
fix veryl-lang/veryl#1574


dalance:
Now creating `std` project by user is not forbidden. So removing `$` causes conflict. I think there are two ways to resolve it.

* replacing `$` to `__`, and forbiding project name with `__`
* forbiding project name with `std`

taichi-ishitani:
I think `std` may be used as a starting word of project name so the 1st choice is better.

Issue #1415, opened 2025-03-21
AXI interface version in std


dalance:
In #1410, `axi_if` covers AXI3 and AXI4.
How we should treat versions?

* `axi_if` covers all versions through parameter or generics
* Split `axi_if` to `axi3_if` and `axi4_if`, and add `axi5_if` in future

janschiefer:
FINALLY I got some time to redo the AXI implementation.

I¨ve implemented the axi interfaces in 2 different ways to choose from --> [](https://github.com/janschiefer/test_axi)

**Variant 1 ( axi_if.veryl && axi_test.veryl ): Implement 2 different axi interfaces for AXI4 and AXI3**
- Pros: 
-- Works with open source tools ( [yosys](https://github.com/YosysHQ/yosys), [sv2v](https://github.com/zachjs/sv2v) )
-- No unused wires
- Cons: Code duplication

**Variant 2 ( axi_if_variant.veryl && axi_test_variant.veryl ): AXI3 & AXI4 uniform implementation using proto package** 
- Pros:
-- More concise, code reusage
- Cons:
-- Incompatible with yosys as well as sv2v
-- Unused wires with width of 1 bit since I see no way to purge them conditionally

Interestingly another bug unmasked:
--> " omit_project_prefix = true " is ignored in interface definitions, but not when the interface is instantiated in a module.

dalance:
Thank you for your work!

I think `axi3_if/axi4_if` is better than generic `axi_if` because AXI5, 6 and so on in future may not suite the current definition of `axi_if`.
But if all parameters are specified individually, the definition of AXI components becomes busy.

```systemverilog
module axi4_slice::<ADDR_W: const, DATA_W_BYTES: const, ID_W: const, AWUSER_W: const, WUSER_W: const, BUSER_W: const, ARUSER_W: const, RUSER_W: const> (
    slv_if: modport axi4_if::<ADDR_W, DATA_W_BYTES, ID_W, AWUSER_W, WUSER_W, BUSER_W, ARUSER_W, RUSER_W>::slave,
    mst_if: modport axi4_if::<ADDR_W, DATA_W_BYTES, ID_W, AWUSER_W, WUSER_W, BUSER_W, ARUSER_W, RUSER_W>::master,
) {}
```

So from this viewpoint, using proto package may be better like below:

```systemverilog
// define generic component
module axi4_slice::<PKG: axi4_prototype> (
    slv_if: modport axi4_if::<PKG>::slave,
[...]

PR #1577, opened 2025-05-02
Implement AXI4, AXI3, AXI4-Lite interfaces and prototypes in veryl_std.

janschiefer:
Implement AXI4, AXI3, AXI4-Lite interfaces and prototypes in veryl_std.

Discussion [here](https://github.com/veryl-lang/veryl/issues/1415).


taichi-ishitani:
I think:

* enum, struct or typedef for axsize, axlen, axburst, axcache, axproto, bresp/rresp should be defined.
* split files for pacakge definition and interface definition

taichi-ishitani:
AXI spec says that AX4 Lite slave components can have ID reflection logic so I think the interface should have awid, bid, arid and rid.

image

taichi-ishitani:
How about using struct for axcache and axprot like this?
https://github.com/taichi-ishitani/tvip-axi/blob/master/src%2Ftvip_axi_types_pkg.sv#L46-L64

Do we need to separate enums for bresp and rresp?

janschiefer:
Converting these consts to packed structs is a very elegant solution.

:heavy_check_mark: Repace constants with packed structs.

I propose to leave bresp_variants and rresp_variants as they are, because if we sometime support extended AXI features like deferred writes or PRI requests rresp and bresp will be 3 bits wide with different meaning between rresp and bresp.

taichi-ishitani:
> if we sometime support extended AXI features like deferred writes or PRI requests rresp and bresp will be 3 bits wide with different meaning between rresp and bresp.

I think we don't need to support such extended AXI as std library because it is no longer AXI.

janschiefer:
:heavy_check_mark: Unify bresp_t (write response) and rresp_t (read response) types into resp_t

dalance:
Thank you for your work!
I'll merge this PR.

Issue #1578, opened 2025-05-02
Interfaces and packages from veryl_std generate inconsistent and invalid names with "$std"


janschiefer:
I updated my [commit](https://github.com/veryl-lang/veryl/pull/1577) and split axi*_if and axi*_pkg.

However another problem unmasked.

Inconsistent interface and package names (partially illegal names) are generated in the System Verilog code. 
You only notice them if you run verilator or a synthesis tool (yosys-slang for example here). Veryl check runs without errors.

Verilator:
```
[DEBUG]    Verilator : %Error: /home/jschiefer/Code/axi_components/dependencies/std/axi_if/axi_pkg.sv:56:9: syntax error, unexpected ${pli-system}, expecting IDENTIFIER or IDENTIFIER-for-cell or IDENTIFIER-for-type or randomize
[DEBUG]    Verilator :    56 | package $std___axi4_pkg__32__4__1__8__8__8__8__8;
[DEBUG]    Verilator :       |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[DEBUG]    Verilator :         ... See the manual at https://verilator.org/verilator_doc.html?v=5.035 for more assistance.
[DEBUG]    Verilator : %Error: /home/jschiefer/Code/axi_components/dependencies/std/axi_if/axi_pkg.sv:87:1: syntax error, unexpected endpackage
[DEBUG]    Verilator :    87 | endpackage
[DEBUG]    Verilator :       | ^~~~~~~~~~
[DEBUG]    Verilator : %Error: /home/jschiefer/Code/axi_components/dependencies/std/axi_if/axi_if.sv:45:11: syntax error, unexpected ${pli-system}, expecting IDENTIFIER or IDENTIFIER-for-cell or IDENTIFIER-for-type or randomize
[DEBUG]    Verilator :    45 | interface $std___axi4_if__$std___axi4_pkg__32__4__1__8__8__8__8__8;
[DEBUG]    Verilator :       |           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #1584, opened 2025-05-03
Remove `$` from emitted component name

taichi-ishitani:
fix veryl-lang/veryl#1578


janschiefer:
cargo clean before push?

Issue #1608, opened 2025-05-08
Shorten mangled name


taichi-ishitani:
Mangled name for a generic component with many generic parameters is difficult to read it because it is very long.
For exaxmple:

Veryl
```systemverilog
package pzcorebus_pkg::<
  PROFILE             : pzcorebus_common_pkg::pzcorebus_profile,
  ID_WIDTH            : u16                                    ,
  ADDRESS_WIDTH       : u16                                    ,
  DATA_WIDTH          : u16                                    ,
  UNIT_DATA_WIDTH     : u16                                    ,
  USE_BYTE_ENABLE     : bool                                   ,
  ATOMIC_COMMAND_WIDTH: u16                                    ,
  MESSAGE_CODE_WIDTH  : u16                                    ,
  RQEUEST_INFO_WIDTH  : u16                                    ,
  RESPONSE_INFO_WIDTH : u16                                    ,
  MAX_DATA_WIDTH      : u16                                    ,
  RESPONSE_BOUNDARY   : u16                                    ,
> for pzcorebus_proto_pkg {
```

Generated SV

```systemverilog
package pzbcm___pzcorebus_pkg__pzcorebus_profile_MEMORY_H__16__32__256__32__true__8__8__0__0__256__4096;
```

I think it is useful to shorten such mangled name.

dalance:
I agree this concept, but I have some consideration points.

* This should be optional, or by default?
* Index based name is unstable, from the point of view of stability, hash based name may be better? (e.g. `prj___PkgA__a138s9fg`)
* In the generated code, the coment which shows the original parameter or long mangled name may be useful

```
// prj::PkgA::<4, 5, 6, 7>
package prj___PkgA__1;
```

taichi-ishitani:
I updated the PR.

* Use hash value instead of instance index
* Print the original instance name as comment

Users can't know the orignal parameters from the shortened generic instace name so I think this feature should be optional.

PR #1614, opened 2025-05-12
Add a build option to shorten mangled name

taichi-ishitani:
close veryl-lang/veryl#1608

Mangled generic instance names will be shortened when the `shoten_mangled_name` option is set.
Example:
https://github.com/veryl-lang/veryl/blob/8a9c670f8af1f4704be3893dad0cf5038faea025/crates/emitter/src/tests.rs#L1474-L1515


dalance:
I changed the title of this PR.
Refer: #1682

PR #1656, opened 2025-05-21
Import slicer module from pzbcm

taichi-ishitani:
This PR is to import slicer module from pzbcm.
https://github.com/pezy-computing/pzbcm/blob/master/pzbcm_slicer/pzbcm_slicer.sv


dalance:
I found typo just after pushing merge button.
So I fixed it directly.

https://github.com/veryl-lang/veryl/commit/3a804d7b3cd25e45a894e6d5643657b08e2d5db8

Issue #1657, opened 2025-05-21
Filelist order is incorrect


taichi-ishitani:
```systemverilog
// a/a.veryl
proto package ProtoPkg {
  const WIDTH: u32;
}

package Pkg::<W: u32> for ProtoPkg {
  const WIDTH: u32 = W;
}
```

```systemverilog
// a/b.veryl
module ModuleA::<PKG: ProtoPkg> {
  let _a: logic<PKG::WIDTH> = 0;
}
```

```systemverilog
// a/c.veryl
module top {
  inst u: ModuleB::<Pkg::<8>>;
}
```

```systemverilog
// b/a.veryl
module ModuleB::<PKG: ProtoPkg> {
  inst u: ModuleA::<PKG>();
}
```

For the above case, its dependency graph is:

```
a/c.veryl -> b/a.veryl -> a/b.veryl -> a/a.veryl
```

Therefore, generaed filelist should be:

```
a/a.sv
a/b.sv
b/a.sv
a/c.sv
```

But actual filelist is:

```
/home/taichi/workspace/temp/veryl/a/b.sv
/home/taichi/workspace/temp/veryl/a/a.sv
/home/taichi/workspace/temp/veryl/b/a.sv
/home/taichi/workspace/temp/veryl/a/c.sv
```

PR #1660, opened 2025-05-22
Insert type dag edge between base and actual generic arg

taichi-ishitani:
fix veryl-lang/veryl#1657

Currently, no dag edge between base symbol and actual arg symbol is inserted if the given path includes references to generic paramters.
This causes #1657.

To fix this case, we need to replace references to generic parameters with actual generic arguments and resolve the path.

### DIFF
--- a/crates/analyzer/src/type_dag.rs
+++ b/crates/analyzer/src/type_dag.rs
@@ -1,6 +1,6 @@
 use crate::AnalyzerError;
 use crate::namespace::Namespace;
-use crate::symbol::{ParameterKind, Symbol, SymbolId, SymbolKind};
+use crate::symbol::{GenericMap, ParameterKind, Symbol, SymbolId, SymbolKind};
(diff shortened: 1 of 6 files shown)


dalance:
Could you add this case to filelist test?

https://github.com/veryl-lang/veryl/blob/cd54151af955db0bf7892cc76ad816bd41f5a421/crates/tests/src/lib.rs#L344
https://github.com/veryl-lang/veryl/tree/master/testcases/filelist

taichi-ishitani:
> Could you add this case to filelist test?
> 
> https://github.com/veryl-lang/veryl/blob/cd54151af955db0bf7892cc76ad816bd41f5a421/crates/tests/src/lib.rs#L344
> 
> 
> https://github.com/veryl-lang/veryl/tree/master/testcases/filelist

Thank you for your feedback.
I added the testcase.

PR #1695, opened 2025-06-04
add const declarations to StatementBlockItems

onsdagens:
This allows the presence of const declarations in statement blocks (e.g. function declarations)

``` 
function foo () -> u32 {
  const bar: u32 = 2;
  return bar;
}
```  

Is there some issue i'm not seeing with something like this?

Also, is there a cleaner way to contribute to the parser? The diff in the generated files is huge.


dalance:
I'll merge this PR.

> Also, is there a cleaner way to contribute to the parser? The diff in the generated files is huge.

It is expected that the diff is huge. No problem.

PR #1718, opened 2025-06-17
Add support for including additional files for tests

onsdagens:
Hi!

This is a PoC implementation of something like what @jbeaurivage mentioned in Discord.

"  From what I understand, veryl test runs the simulations from a temp directory which is cleaned up after the run (at least for verilator, not sure about other simulation frameworks). Seems like there is no mechanism to include extra files to the temp directory so that they can be referenced in the simulation, is that correct? "

This adds additional option in Veryl.toml

```
[test.verilator]
include_files = [string]
```

The provided files will be copied into the temp directory created for the test bench. My personal use case for this is `$readmemh()` based memory population, which requires a source `.mem` file.

Current implementation is limited to an array of file paths, and to Verilator (the way you want to do this for vcd or Vivado may differ, and i'm not familiar enough with those to do it), one could also think of allowing the inclusion of entire directories.

I've not used `miette` before, so i'm unsure of the idiomatic way of creating errors, i've used miette::bail!(), with a hopefully helpful message, let me know if this needs changing.

### DIFF
--- a/crates/metadata/src/test.rs
+++ b/crates/metadata/src/test.rs
@@ -16,6 +16,8 @@ pub struct Test {
     pub waveform_target: WaveFormTarget,
     #[serde(default)]
     pub waveform_format: WaveFormFormat,
+    #[serde(default)]
+    pub include_files: Vec<PathBuf>,
 }
 
 #[derive(Clone, Copy, Debug, Default, Serialize, Deserialize, PartialEq, Eq)]
--- a/crates/veryl/src/runner/cocotb.rs
+++ b/crates/veryl/src/runner/cocotb.rs
@@ -1,6 +1,6 @@
 use crate::runner::{Runner, copy_wave};
 use futures::prelude::*;
-use log::{error, info};
+use log::{error, info, warn};
 use miette::{IntoDiagnostic, Result, WrapErr};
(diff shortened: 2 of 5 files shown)


dalance:
I think this option is not simulator-specific, so it should be in `[test]`.
If you can't test these simulator, how about using `unimplemented!` to indicate it is not supported yet.
Users who use these simulator will find the panic and can implement it.

onsdagens:
I imagine in some cases you may want to include some files for let's say vivado but not verilator (i do not have a specific use case for this, it's a gut feeling). How about a global list under `[test]` and backend specific lists as implemented?

dalance:
I agree such possibility, but I think we can add tool specific list after finding actual use cases.

onsdagens:
I've moved the setting out to global `[test]` section as mentioned, and implemented support for Vivado. 

For unsupported backends, i've opted to warn instead of panic, since otherwise, having any includes in your project would mean you cannot use unsupported test backends at all (even if the includes may be unneeded for that backend). Let me know what you think about this, i'm open to changing it to `unimplemented!()`, as mentioned before, since as you said we don't really have a use case for different include lists between backends.

dalance:
Thanks!
I think warning message is better than panic.

PR #1729, opened 2025-06-24
Add support for additional files with cocotb runner

jbeaurivage:
Following #1718, add support for file includes with `cocotb` runner.

One caveat is that the included files are copied into the *parent* of the current working directory of the test script.


dalance:
Looks good!

Issue #1735, opened 2025-06-26
Emit pacakge item name without package prefix


taichi-ishitani:
Veryl emits illegal SV from Veryl code with following condition.

* Refer package items with **pacakge prefix** in parameter/port list
* Import the above pacakge item in the module body

Example

Veryl
```systemverilog
package Pkg {
    const W: u32 = 8;
}

module ModuleA (
    i_d: input  logic<Pkg::W>,
    o_d: output logic<Pkg::W>,
) {
    import Pkg::W;
    assign o_d = i_d;
}
```

Generated SV
```systemverilog
package project_Pkg;
    localparam int unsigned W = 8;
endpackage

module project_ModuleA (
    input  var logic [W-1:0] i_d,
    output var logic [W-1:0] o_d
);
    import project_Pkg::W;
    always_comb o_d = i_d;
endmodule
```

Reference to `W` in the port list should be `project_Pkg::W` because `project_Pkg` has not been imported at this line.

taichi-ishitani:
Currently, `import` declarations in module body does not affect module body only but also parameter list and port list.
I think `import` declarations should be emitted into module header but not module body.

PR #1737, opened 2025-06-27
Move import declarations to component header

taichi-ishitani:
close veryl-lang/veryl#1735

For Veryl, import declarations do not affect only component body but also component header.
Therefore, import declarations should be emitted to component header.


dalance:
Could you split the differences caused by clippy update to another PR?

taichi-ishitani:
> Could you split the differences caused by clippy update to another PR?

I removed changes fixing clippy errors from this PR.

Issue #1786, opened 2025-07-12
Not add prefix/suffix to hierarchical referenced reset signal


taichi-ishitani:
Prefix/suffix is not added to reset signal referenced via hierarchical reference in `always_comb` block.

Veryl
```systemverilog
interface FooIF {
  var rst: reset;
}

module Bar (
  i_rst: input reset
) {
  inst foo_if: FooIF;
  always_comb {
    foo_if.rst = i_rst;
  }
}
```

Generated SV
```systemverilog
interface veryl_testcase_FooIF;
    logic rst_n;
endinterface

module veryl_testcase_Bar (
    input var logic i_rst_n
);
    veryl_testcase_FooIF foo_if ();
    always_comb begin
        foo_if.rst = i_rst_n;
    end
endmodule
```

PR #1787, opened 2025-07-12
fix issue that prefix/suffix are not added to clock/reset signal in expression identifier

taichi-ishitani:
fix veryl-lang/veryl#1786


dalance:
Does this PR include generated parser changes caused by parol update?
If so, could you split it into another PR?

taichi-ishitani:
> If so, could you split it into another PR?

Ok, I will remove the parser update from this PR.
#1783 and #1778 include the parser update.

Issue #1769, opened 2025-07-08
Remove `ClockDomain` scanner state


taichi-ishitani:
The `ClockDomain` scanner state has been introduced to distinguish simple `'` token and other tokens starting with `'`.

Parol v3 supports the longest match rule so such division of state is not necessary.
https://github.com/jsinger67/parol/issues/715#issuecomment-3048079782

refs: #1751

PR #1817, opened 2025-07-29
Remove `ClockDomain` scanner state

taichi-ishitani:
close veryl-lang/veryl#1769


dalance:
Nice! Parse performance was improved about 7%.

Issue #1838, opened 2025-08-06
Can't use expression identifier scoped by generic package instance as generic arg


taichi-ishitani:
Veryl compiler reprots `unresolvable_generic_argument` and `mismatch_type` for expression identifier which is scoped by generic package instance and used as a generic arg.

Veryl
```systemverilog
package PkgA::<V: u32> {
    struct Foo {
        foo: u32,
    }
    const FOO: Foo = Foo'{ foo: V };
}

package PkgB::<V: u32> {
    const BAR: u32 = V;
}

module ModuleA {
    import PkgB::<PkgA::<32>::FOO.foo>::*;
}
```

```
]8;;https://doc.veryl-lang.org/book/07_appendix/02_semantic_error.html#unresolvable_generic_argument\unresolvable_generic_argument (link)]8;;\

  💥 FOO can't be resolved from the definition of generics
    ╭─[:8:9]
  7 │ 
  8 │ package PkgB::<V: u32> {
    ·         ──┬─
    ·           ╰── Definition
  9 │     const BAR: u32 = V;
 10 │ }
 11 │ 
 12 │ module ModuleA {
 13 │     import PkgB::<PkgA::<32>::FOO.foo>::*;
    ·                   ─────────┬─────────
    ·                            ╰── Error location
 14 │ }
    ╰────
  help: 
]8;;https://doc.veryl-lang.org/book/07_appendix/02_semantic_error.html#mismatch_type\mismatch_type (link)]8;;\

  💥 "__PkgA__32 FOO foo " is expected to "u32", but it is "identifier"
    ╭─[:13:19]
 12 │ module ModuleA {
 13 │     import PkgB::<PkgA::<32>::FOO.foo>::*;
    ·                   ─────────┬─────────
    ·                            ╰── Error location
 14 │ }
    ╰────
  help: 
```

taichi-ishitani:
For the first error, it is reported when resolving `PkgB`'s generic parameter `V`.

It is replaced with the actual generic parameter `PkgA::<32>::FOO.foo` at here.
https://github.com/veryl-lang/veryl/blob/e22a7bb2e4525f5a5667bb735e5b1ed380a40004/crates/analyzer/src/reference_table.rs#L227
Due to this, result of `prefix_len` is wrong and then symbol resolutin for `PkgA::<32>::FOO.foo` is failed.
[...]

PR #1840, opened 2025-08-07
Fix `unresolvable_generic_argument` and `mismatch_type` errors for expression identifier used as generic arg

taichi-ishitani:
fix veryl-lang/veryl#1838


taichi-ishitani:
This PR is merged with #1841 .

Issue #1842, opened 2025-08-07
Number of printed error messages


taichi-ishitani:
Currently, all analyze errors are printed so it is hard to see messages at the beginning when a lot of errors is reported.
It is useful to allow users to configure number of printed error messages.

taichi-ishitani:
Introduce the `error_count_limit` field to the `build` option for this purpose.

PR #1864, opened 2025-08-15
Add `error_count_limit` build option

taichi-ishitani:
close veryl-lang/veryl#1842


crates/veryl/src/cmd_check.rs
```diff
@@ -44,6 +57,10 @@ impl CheckError {
             Err(self.into())
         }
     }
+
+    fn is_error(error: &AnalyzerError) -> bool {
```
dalance, line 61: It may be better that this function is method of `AnalyzerError`.

crates/veryl/src/cmd_check.rs
```diff
@@ -12,25 +12,38 @@ pub struct CmdCheck {
...
+    pub fn new(error_count_limit: u32) -> Self {
+        Self {
+            related: Vec::new(),
+            error_count: 0,
+            error_count_limit,
+        }
+    }
+
+    pub fn append(mut self, x: &Vec<AnalyzerError>) -> Self {
+        for error in x {
```
dalance, line 34: To avoid `clone`, you can use `drain` like below:

```rust
for error in x.drain(0..) {
}
```

`drain` takes `&mut Vec`, and the `error` can be directly added to `self.related` because it is moved from `x`.

taichi-ishitani:
Thank you for your feedbacks and I fixed them.

Issue #1898, opened 2025-09-11
Unexpected `mismatch_generics_arity` error


taichi-ishitani:
Veryl compiler reports `mismatch_generics_arity` error for the code below unexpectedlly.

```systemverilog
package PkgA::<V: u32> {
  const A: u32 = V;
}

module ModuleA {
  import PkgA::<32>::*;

  function get_v::<V: u32> -> u32 {
    return V;
  }

  let _a: u32 = get_v::<A>();
}
```

```
]8;;https://doc.veryl-lang.org/book/07_appendix/02_semantic_error.html#mismatch_generics_arity\mismatch_generics_arity (link)]8;;\

  💥 generics "PkgA" has 1 generic arguments, but 0 arguments are supplied
    ╭─[:12:25]
 11 │ 
 12 │   let _a: u32 = get_v::<A>();
    ·                         ┬
    ·                         ╰── Error location
 13 │ }
    ╰────
  help: fix generics arguments
```

taichi-ishitani:
The code below makes Veryl compiler panicked.
It should also be fixed.

```systemverilog
package PkgA::<V: u32> {
  const A: u32 = V;
}

module ModuleA {
  import PkgA::<32>::*;

//  function get_v::<V: u32> -> u32 {
//    return V;
//  }

  let _a: u32 = A;
}
```

PR #1899, opened 2025-09-11
fix unexpected `mismatch_generics_arity` error

taichi-ishitani:
fix veryl-lang/veryl#1898
fix veryl-lang/veryl#1900

This error is reported when resolving generic arg imorted from generic package.

Path to an imported symbol is appended with the package where the symbol belongs to. However, expanded path does not include generic args even if the parent package is a generic package.
https://github.com/veryl-lang/veryl/blob/d55b0f4003decf3fdc91497cccc15c9e234f2e2f/crates/analyzer/src/symbol_path.rs#L565

Due to this, unexpected `mismatch_generics_arity` error is reported.


taichi-ishitani:
This PR also fixed #1900 .

PR #1959, opened 2025-10-10
Make cocotb version check stricter

omasanori:
This change should be compatible with Python 3.8 or later.

### DIFF
--- a/crates/veryl/src/runner/cocotb.rs
+++ b/crates/veryl/src/runner/cocotb.rs
@@ -194,9 +194,10 @@ impl Runner for Cocotb {
         let runner_path = temp_dir.path().join("runner.py");
         let runner_text = format!(
             r#"
-import cocotb
+from importlib.metadata import version
 
-if cocotb.__version__.startswith("2"):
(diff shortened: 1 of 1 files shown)


dalance:
Is `veryl test` failed if user doesn't install `packaging`?
If so, it seems to be inconvenience.

Is it possible to do a stricter check using str operation?

omasanori:
I made a mistake anyways...

> Is veryl test failed if user doesn't install packaging?

With this change, yes.

> Is it possible to do a stricter check using str operation?

OK, I will try.

omasanori:
However, it will not be simple str operation but a kind of parsing. (Consider "10" is lexicographically earlier than "2".)

dalance:
I agree.
I think it is acceptable that only "1.9.x" and "2.x" (or "2.0.x") are supported to simplify version detection logic.

omasanori:
Ah yes, it seems reasonable. Let me revise this PR.

omasanori:
@dalance I think it becomes ready for your review.

dalance:
Looks good!

PR #1960, opened 2025-10-10
Support Python 3.6+

omasanori:
RHEL 8.x and its derivatives will be supported until May 2029, which uses Python 3.6 by default.

cocotb 2.0 supports Python 3.6, so it seems reasonable to give it a chance.


dalance:
I think this change seems to be no problem.
Do you have any remaining issue?

omasanori:
Sorry for my inactivity. I will test this patch on AlmaLinux 8 and then turn it ready for review.

omasanori:
I have installed Python 3.6.15 locally and confirmed that it works as expected.

dalance:
Thanks!
I'll merge this PR.

Issue #1990, opened 2025-10-20
`type` generic param can't be used for LHS value of const type declaration


taichi-ishitani:
`invalid_factor` error will be reported for the code below.

```systemverilog
proto package proto_pkg {
    const T: type;
}
package pkg::<t: type = bool> {
    const T: type = t;
}
```

```console
]8;;https://doc.veryl-lang.org/book/07_appendix/02_semantic_error.html#invalid_factor\invalid_factor (link)]8;;\

  💥 t of kind "generic parameter" cannot be used as a factor in an expression
   ╭─[:5:15]
 4 │ 
 5 │ package pkg::<t: type = bool> {
   ·               ┬
   ·               ╰── Error location
 6 │     const T: type = t;
   ╰────
  help: remove generic parameter from expression
```

taichi-ishitani:
Currenlty, type factors can be used in any expressions like below. It also shoul be fixed.

<img width="1358" height="263" alt="Image" src="https://github.com/user-attachments/assets/ffe9049e-3945-4bfe-828c-f215c1f65077" />

PR #1999, opened 2025-10-23
Implement type expression check

taichi-ishitani:
fix veryl-lang/veryl#1990

The `Evaluator` does not support `type` expressions and it returns error result for them.
This causes #1990.

To fix this error, implement evaluation of type expression and some check for thier results.


taichi-ishitani:
@dalance ,

Thank you for your review.
I fixed feedbacks from you.

Issue #1979, opened 2025-10-16
Option to emit `case` statement without `inside` keyword


taichi-ishitani:
Currenlty, `case` statement is emitted into `case inside` statement.
However, funcations that include `case inside` statement can't be put RHS of parameter delcarations.

```
Error-[NYI] Not Yet Implemented
/home/ishitani/workspace/pzdma/hw/sim/pzdma_256b_dual_engines_rtl/sim_binary/dependencies/pzdma/hw/rtl/pzdma/pzdma_source_bus_adapter.sv, 39
  Feature is not yet supported: inside operator is not supported by constant
  function evaluation.
  Source info: case (max_burst_size) inside
  pzdma_burst_length_BURST_LENGTH_32:
```

To resolve this, a new option to select how to eimit `case` statement is needed.

dalance:
I think there are two candidates.

* Veryl compiler detects whteher each `case` statement need `inside`, and suppresses unnecessary `inside`.
* User specify that each `case` statement require `inside` or not by atttibure.

taichi-ishitani:
I prefer the 1st candidate.
To consider this, we need to consider how to evaluate case condition items.

If a parameter used as a case condition includes x/z bits, such case statement can't be emitted into a simple case statement.
To simplify evaluation of parameter value, it is reasonable that paramter whoes type is 4 state type is not treated as fixed value even if  it does not include x/z bits.

PR #2005, opened 2025-10-28
Introduce IR

### LINKED ISSUE #1719: `const` variables do not trigger CDC errors
```systemverilog
module cdc (
    i_a: input  'a logic,
    o_b: output 'b logic,
) {
    const inv: logic = ~i_a;
    // assign inv = ~i_a;
    assign o_b = inv;
}
```

Expected behavior: `const inv: logic = ~i_a;` should trigger a CDC error, since its type is not annotated with clock domain `a`.
As follow-up, `assign o_b = inv;` should trigger a CDC error.

Actual behavior: this code checks and transpiles without error or warning.

Version: `veryl 0.16.1-nightly (b259fbd 2025-06-17)`

### LINKED ISSUE #1847: Introduce intermediate representation (IR)
Now all operations are based on syntax tree.
This is not probably suitable for some stages of analyzer.
To reduce implementation complexity, I'll consider about introducing IR in this issue.

### LINKED ISSUE #2046: Should operand of logical operation be 1-bit value?
#2005 can introduce more accurate type checking.
So should we introduce to check whether operand of logical operation is 1bit value?

If so, the following is all failed, because `1` is 32-bit value.
This may be too restricted?

```systemverilog
always_comb {
    a = 1 && 1;
    a = !1;
    if 1 {
    }
}
```

### LINKED ISSUE #2052: Move analyzer pass2 to IR constructor
IR construction is planned to occur after the pass2 stage of the analyzer.

The pass2 stage performs semantic checks by traversing the syntax tree using the handler mechanism, visiting each node only once. However, IR construction requires visiting the same nodes multiple times to handle generate-for, instantiation, etc. Therefore, pass2 and IR construction cannot be merged.

On the other hand, one of the pass2 checkers, `check_expression`, recursively calls pass2 internally to account for parameter overrides during instantiation — which is exactly the same operation performed during IR construction.

Therefore, it is reasonable to move the entire pass2 semantic checking into the IR construction phase. This will naturally enable proper handling of parameter overrides in several checkers that currently do not account for them.

### LINKED ISSUE #2079: Not detect multiple driver error
Currenlty, veryl compiler does not detect `multiple driver` error.
As a result, invalid SV is generated.

Example:

```systemverilog
module ModuleA {
  var a: logic;

  always_comb {
    a = 0;
  }
  always_comb {
    a = 1; // Illegal assignment but no error is reported
  }
}
```

### LINKED ISSUE #2110: Can't use generic param as a casting type
Veryl reports `mismatch_type` error for the code below.

```systemverilog
module ModuleA::<WIDTH: u32> {
    let _a: logic<WIDTH> = 0 as WIDTH;
}
```

```console
]8;;https://doc.veryl-lang.org/book/07_appendix/02_semantic_error.html#mismatch_type\mismatch_type (link)]8;;\

  💥 "WIDTH" is expected to "enum or union or struct", but it is "generic
  │ parameter"
   ╭─[:2:33]
 1 │ module ModuleA::<WIDTH: u32, TYPE: type> {
 2 │     let _a: logic<WIDTH> = 0 as WIDTH;
   ·                                 ──┬──
   ·                                   ╰── Error location
 3 │     let _b: TYPE         = 0 as TYPE ;
   ╰────
  help: 
```

### LINKED ISSUE #2123: Wrong bit position calcuration of `step` part select
Veryl reports `invalid_select` error due to wrong bit position calcuration of `step` part select.
For the example below, `_a[1 step 16]` is eq `_a[1*16+:16]` and this is valid range. However, the error is reported.

```systemverilog
module module_a {
    let _a: logic<32> = 0;
    let _b: logic<16> = _a[1 step 16];
}
```

```console
]8;;https://doc.veryl-lang.org/book/07_appendix/02_semantic_error.html#invalid_select\invalid_select (link)]8;;\

  ⚠️ invalid select caused by out of range [32:16] > 32
   ╭─[:3:25]
 2 │     let _a: logic<32> = 0;
 3 │     let _b: logic<16> = _a[1 step 16];
   ·                         ──────┬──────
   ·                               ╰── Error location
 4 │ }
   ╰────
  help: 
```

### LINKED ISSUE #2133: Constant expression can not be used in reset block
I want to use a constant expression as a reset value, but currently it results in an error:

```veryl
module reset_delay::<N: u8> ( ... ) {
  var clk_delay: bit<5>;
  always_ff (clk, rst_n) {
    if_reset {
      clk_delay = (1 << (N & 5'h1f)) - 1;
    } else {
      clk_delay = clk_delay >> 1;
    }
  }
```

The error message is "Reset-value cannot be used because it is not evaluable at elaboration time".

The error is vanished when I extract the RHS value as a const value:

```veryl
module reset_delay::<N: u8> ( ... ) {
  const DELAY_VAL: bit<5> = (1 << (N & 5'h1f)) - 1;
  var clk_delay: bit<5>;
  always_ff (clk, rst_n) {
    if_reset {
      clk_delay = RESET_VAL;
    } else {
      clk_delay = clk_delay >> 1;
    }
  }
```

dalance:
Closes #1847
Closes #2046
Closes #2079 
Closes #2110
Closes #2123 
Closes #2052
Closes #1719
Closes #2133


taichi-ishitani:
To implement #1979, we need to check if all case conditions are typed with 2-state types.
Can IR check this?

dalance:
> To implement #1979, we need to check if all case conditions are typed with 2-state types. Can IR check this?

Yes.

taichi-ishitani:
This PR will fix #2110.

PR #2030, opened 2025-11-05
Add build and verilator outputs to .gitignore template

mchoo7:


dalance:
The `/target` entry in gitignore of this repo come from `cargo new` of Rust.
I investigated the reason why trailing slash doesn't exist and found the following PR.

https://github.com/rust-lang/cargo/pull/4944

So I think it is better that the trailing slashes of  `/target/` and `/dependencies/` are removed.
This is because these path may be symlink created by users.
I'm sorry again, but could you remove them?

mchoo7:
> The `/target` entry in gitignore of this repo come from `cargo new` of Rust. I investigated the reason why trailing slash doesn't exist and found the following PR.
> 
> [rust-lang/cargo#4944](https://github.com/rust-lang/cargo/pull/4944)
> 
> So I think it is better that the trailing slashes of `/target/` and `/dependencies/` are removed. This is because these path may be symlink created by users. I'm sorry again, but could you remove them?

Fixed!

Same for `/obj_dir/` (generated by verilator)?

dalance:
> Same for `/obj_dir/` (generated by verilator)?

I feel either is fine because tool generated paths are probably real directory.

dalance:
I think this PR is ready to merge.
I'll merge after passing CI.

Issue #1876, opened 2025-08-25
Add default implementation of `std::synchonizer`


taichi-ishitani:
Add the module like below link as the defualt implementation of `$std::synchonizer`.
https://github.com/pezy-computing/pzbcm/blob/master/pzbcm_synchronizer/pzbcm_synchronizer.sv

PR #2036, opened 2025-11-06
Add basic synchronizer implementation

taichi-ishitani:
close veryl-lang/veryl#1876


crates/std/veryl/src/synchronizer/synchronizer_basic.veryl
```diff
@@ -0,0 +1,93 @@
+/// Basic synchronizer implementation using multi-bit FF scheme
```
dalance: I think "multi-bit FF" generally means multi-bit FF cell.

https://www.techsimplifiedtv.in/2022/12/what-is-multi-bit-flip-flop-mbff-in-vlsi.html

So "Basic synchronizer implementation using single-bit FF scheme" is correct because this implementation doesn't use MBFF cell.

taichi-ishitani: Thanks for your review.
I modified the comment.

PR #2061, opened 2025-11-21
Simplify generic reference resolution

### LINKED ISSUE #2056: Unknown member error is not detected for const struct imported from proto package
Following code has an unknown member reference but it is not reported.

```systemverilog
proto package a_proto_pkg {
    struct foo_struct {
        foo: u32,
    }

    const FOO: foo_struct;
}

package a_pkg::<V: u32> for a_proto_pkg {
    struct foo_struct {
        foo: u32,
    }

    const FOO: foo_struct = foo_struct'{ foo: V };
}

module b_module::<PKG: a_proto_pkg> #(
    const BAR: u32 = FOO.bar, // bar is unknown member
) {
    import PKG::*;
}

alias package A = a_pkg::<32>;
alias module  B = b_module::<A>;
```

taichi-ishitani:
fix #2056 

Currenlty, generic referecnes are resolved when all generic args are applied.
However, generic parameters are typed by `proto` type so symbol resulutions can be done with generic parameters before appplying generic args.

This PR is to simplify generic reference resolution:

* Resolving symbol path even if it includes unresolived generic parameters
    * symbol path resolution is done with `proto` boundary
* Split resolving symbol path and inserting generic instance
    * generic instance is inserted recursibly when all generic args are applied

Changes in this PR also fixes #2056.

### DIFF
--- a/crates/analyzer/src/handlers/create_symbol_table.rs
+++ b/crates/analyzer/src/handlers/create_symbol_table.rs
@@ -42,30 +42,19 @@ use veryl_parser::veryl_walker::{Handler, HandlerPoint};
 #[derive(Default)]
 struct GenericContext {
     parameters: Vec<Vec<SymbolId>>,
-    references: Vec<Vec<GenericSymbolPath>>,
 }
 
 impl GenericContext {
     pub fn push(&mut self) {
         self.parameters.push(Vec::new());
-        self.references.push(Vec::new());
     }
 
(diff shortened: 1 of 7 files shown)


taichi-ishitani:
This change also reduces run time.
Following results are build time of our DMAC core project.

* nightly
  ```
  real    3m7.665s
  user    0m42.177s
  sys     2m24.502s
  ```
* with this PR
  ```
  real    0m2.406s
  user    0m2.330s
  sys     0m0.071s
  ```

Issue #2073, opened 2025-11-26
Generic instance referenced as LHS of alias declaration is not inserted


taichi-ishitani:
Generic instances referenced as the LHS of an alias declaration may not be inserted.
Generic arg of such instance is defined in the same namespace of the alias.
As a result, wrong SV for such generic instance is generated.

Example:

* a.veryl
  ```systemverilog
  package a_pkg::<V: u32> {
      const A: u32 = V;
  }
  ```
* b.veryl
  ```systemverilog
  package b_pkg {
      const B: u32 = 32;
      alias package pkg = a_pkg::<B>;
  }
  ```

Output

* a.sv
  ```systemverilog
  package veryl_testcase___a_pkg__B;
      localparam int unsigned A = ;
  endpackage
  //# sourceMappingURL=a.sv.map
  ```

taichi-ishitani:
There is a cyclic dependency between `a_pkg` and `b_pkg` because `b_pkg` depends on `a_pkg` and `a_pkg` depends on `B`.
However, no error is reported.

taichi-ishitani:
~~`b_pkg::B` can't be referenced from `a_pkg` so `unresolvable_generic_argument` error should be reported for this case.~~

PR #2074, opened 2025-11-26
Treat generic arg defined in namespace same with the namespace of generic reference as unresolvable generic arg

taichi-ishitani:
fix veryl-lang/veryl#2073

```systemverilog
package PkgA::<V: u32> {
    const A: u32 = V;
}
package PkgB {
    const B: u32 = 32;
    const A: u32 = PkgA::<B>::A;
}
```

For the above example, `PkgA` depends on `PkgB::B` and `PkgB` depends on `PkgA` so these packages are depended cyclicly.
However, no errors are reported.

This PR is to make generic arg `B` to be treated unresolved generic arg.

### DIFF
--- a/crates/analyzer/src/handlers/check_type.rs
+++ b/crates/analyzer/src/handlers/check_type.rs
@@ -49,8 +49,12 @@ impl Handler for CheckType {
     }
 }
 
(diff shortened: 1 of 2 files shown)


taichi-ishitani:
Closing this PR because this case should be reproted as `cyclelic dependecy` error.

PR #2078, opened 2025-11-27
Fix issue that cyclic dependency in generic reference may not be detected

taichi-ishitani:
fix veryl-lang/veryl#2073

```systemverilog
package PkgA::<V: u32> {
    const A: u32 = V;
}
package PkgB {
    const B: u32 = 32;
    const A: u32 = PkgA::<B>::A;
}
```

For the above example, `PkgA` depends on `PkgB::B` and `PkgB` depends on `PkgA` so these packages are depended cyclicly.
`B` is not visible from `PkgA` because `B` has no package prefix. Due to this, cyclic dependency between these is not tested.
https://github.com/veryl-lang/veryl/blob/8c6a0e17c1a3ca2f0be377b00adeb1e053c8d73f/crates/analyzer/src/type_dag.rs#L219

To fix this error, we need to add the package prefix to `B` to make `B` visible from `PkgA`.
Then dependency check between `PkgA` and `B` will be tested.

### DIFF
--- a/crates/analyzer/src/symbol_path.rs
+++ b/crates/analyzer/src/symbol_path.rs
@@ -553,6 +553,14 @@ impl GenericSymbolPath {
         path
     }
 
+    pub fn file_path(&self) -> Option<PathId> {
+        if let TokenSource::File { path, .. } = self.range.beg.source {
+            Some(path)
+        } else {
+            None
+        }
+    }
+
     pub fn unalias(&mut self) {
         if !self.is_resolvable() {
             return;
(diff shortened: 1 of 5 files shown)


crates/analyzer/src/symbol_path.rs
```diff
@@ -647,35 +655,6 @@ impl GenericSymbolPath {
         false
     }
 
-    pub fn may_be_generic_reference(&self) -> bool {
```
taichi-ishitani: Remove this function because it is unused.

PR #2182, opened 2026-01-27
BugFix: Keyword completion now works as expected in helix editor

eirikki931:
When using helix editor, keyword completion did not replace the existing text. Resulting in completions like if -> iiff.
Changed behavior to insert text instead of edit text, like the other completion functions in server.rs

Please note that I have only tested the change in Helix Editor.


dalance:
I checked neovim and VSCode, and both of them work fine.
I'll merge this PR.

PR #2197, opened 2026-02-03
Add native simulator engine v2

### LINKED ISSUE #1287: Native simulator
Now all symbol information is already gathered, clock / reset is identified syntactically, and expression evaluator is ready.
So I think implementing Veryl native simulator is not so difficult.

The aim of the simulator is not simulation performance, but quick response.
All other solutions like Verilator and commercial simulator take long   compilation time. 
If simulation of the first one cycle complete in a second, it is probably valuable.

dalance:
Closes #1287


omasanori:
Fantastic! What happens if simulated modules contain `$sv::` or `embed (inline) sv{{{ ... }}}`?

dalance:
> Fantastic! What happens if simulated modules contain `$sv::` or `embed (inline) sv{{{ ... }}}`?

Unfortunately, code including SV will not be supported.

omasanori:
I see, thanks for your explanation.

PR #2244, opened 2026-02-23
Changed version to optional in project metadata

### LINKED ISSUE #2113: `Veryl.toml` without `version` field
I put `Veryl.toml` file to each simulation directories to vary generic args given to the DUT.
Example: https://github.com/taichi-ishitani/rice/blob/master/sim/core/sim_binary/Veryl.toml
For this usage, the `version` info is unnecessary.

However, the `version` field is a mandatly feild so the `MetadataError` error is reported if `Veryl.toml` does not include this field.

To suppor the above usage:
* Change the `version` field to an optional field
* Check existence of this field only when `veryl publish` is executed

prokie:
Fix #2113 

- Changed version to optional in project metadata.

### DIFF
--- a/crates/metadata/src/metadata.rs
+++ b/crates/metadata/src/metadata.rs
@@ -160,15 +160,18 @@ impl Metadata {
             return Err(MetadataError::ModifiedProject(prj_path.to_path_buf()));
         }
 
+        let version = self
+            .project
+            .version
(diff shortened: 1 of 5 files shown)


dalance:
Thanks!
I think the review by Copilot is no problem, so I'll merge this PR.

PR #2253, opened 2026-02-24
fixes incorrect `context_width` propagation in concatenation constant folding

tignear:
There is an issue in the analyzer's implementation of constant folding for concatenations (`{}`). Currently, the expressions within a concatenation are being incorrectly influenced by the `context_width`.

According to IEEE 1800-2023, Section 11.6, Table 11-21 Bit lengths resulting from self-determined expressions, concatenations are defined such that "all operands are self-determined." This implies that the width must not be inferred from the left-hand side (LHS) or the surrounding context. This behavior is also clearly demonstrated in Section 11.6.3 Example of self-determined expressions.

Even if we were not strictly following the standard, extending all operands to the width inferred from the context leads to obviously unintended behavior, as shown in the examples below.

### Example

In the following Veryl code, the addition `8'hff + 8'h1` inside the concatenation should be evaluated as a self-determined 8-bit expression (resulting in `8'h00`).

```veryl
module Top (
    o: output logic<16>,
    o2: output logic<16>,
    o3: output logic<16>,
) {
    // 8'hff + 8'h1 = 8'h00 (self-determined)
    assign o = {8'hff + 8'h1}; 
    assign o2 = 8'hff + 8'h1; 
    assign o3 = {8'hf0, 8'hff + 8'h1};
}
```
### Actual
The analyzer incorrectly propagates the 16-bit context_width from the LHS into the concatenation, causing the addition to be treated as a 16-bit operation (`16'h0100`) and the concatenation width to be calculated incorrectly.
```
module Top {
  output var0(o): logic<16> = 16'hxxxx;
  output var1(o2): logic<16> = 16'hxxxx;
  output var2(o3): logic<16> = 16'hxxxx;

  comb {
    var0 = 16'h0100; // Incorrect
  }
  comb {
    var1 = 16'h0100; 
  }
  comb {
    var2 = 24'hf00100; // Incorrect
  }
}
```
### Expected
The `context_width` should not propagate through `Expression::Concatenation`. The addition remains 8-bit, and the final concatenation result reflects this.
```
module Top {
  output var0(o): logic<16> = 16'hxxxx;
  output var1(o2): logic<16> = 16'hxxxx;
  output var2(o3): logic<16> = 16'hxxxx;

  comb {
    var0 = 8'h00; // Correct: Self-determined 8-bit result
  }
  comb {
    var1 = 16'h0100; 
  }
  comb {
    var2 = 16'hf000; // Correct
  }
}
```

### DIFF
--- a/crates/analyzer/src/ir/tests.rs
+++ b/crates/analyzer/src/ir/tests.rs
@@ -1838,26 +1838,20 @@ fn concat() {
   module Top (
     o: output logic<16>,
     o2: output logic<16>,
-    o3: output logic<16>,
 ) {
     assign o = {8'hff + 8'h1}; 
-    assign o2 = 8'hff + 8'h1; 
-    assign o3 = {8'hf0, 8'hff + 8'h1};
+    assign o2 = {8'hf0, 8'hff + 8'h1};
 }
     "#;
     let exp = r#"module Top {
   output var0(o): logic<16> = 16'hxxxx;
   output var1(o2): logic<16> = 16'hxxxx;
-  output var2(o3): logic<16> = 16'hxxxx;
 
   comb {
     var0 = 8'h00;
   }
   comb {
-    var1 = 16'h0100;
-  }
-  comb {
-    var2 = 16'hf000;
+    var1 = 16'hf000;
   }
 }
 "#;


dalance:
I think the fix is correct, so could you add an unit test to prevent regression to `crates/analyzer/src/ir/tests.rs`?
Your example and expected code is probably sufficient.

PR #2264, opened 2026-02-28
fix(analyzer): correct EqWildcard/NeWildcard

tignear:
Fixes wildcard equality (`==?`/`!=?`) evaluation to comply with IEEE 1800-2023 Section 11.4.6.             
                                                                          
The spec defines three possible outcomes                                                                                                                                                                 
> If the relation is true, the operator yields a `1'b1`. If the relation is false, it yields a `1'b0`. If the relation is unknown, it yields `1'bx`.

The previous implementation could not distinguish `false` from `unknown`. When the LHS contained any X bits at non-wildcard positions, the result was always `1'bx` — even when a definite mismatch at other non-wildcard positions made the relation definitively false.

This only occurred when both conditions were met simultaneously.
1. The LHS has X/Z at a non-wildcard position (would normally make the result unknown)
2. Another non-wildcard position has a definite value mismatch (should make the result false)

  | Expression | Before | After | IEEE 1800 outcome |
  |---|---|---|---|
  | `8'hx3 ==? 8'h30` | `1'bx` | `1'b0` | false: `3 != 0` at non-wildcard position |
  | `8'hx3 !=? 8'h30` | `1'bx` | `1'b1` | true: definite mismatch, inverted |
  | `8'hx0 ==? 8'h30` | `1'bx` | `1'bx` | unknown: no definite mismatch (unchanged) |

### DIFF
--- a/crates/analyzer/src/ir/op.rs
+++ b/crates/analyzer/src/ir/op.rs
@@ -708,25 +708,29 @@ impl Op {
                 let x = x.expand(xy_width, false);
                 let y = y.expand(xy_width, false);
 
-                let (is_zero, is_x) = match (x.as_ref(), y.as_ref()) {
+                let (is_mismatch, is_x) = match (x.as_ref(), y.as_ref()) {
                     (Value::U64(x), Value::U64(y)) => {
-                        let wildcard = !y.mask_xz;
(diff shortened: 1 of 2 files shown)


dalance:
Looks good!
I'll merge this PR.

PR #2265, opened 2026-02-28
prevent keywords from being highlighted as port references

### LINKED ISSUE #1800: Not highlight `always_ff` keyword
VS code plugin does not highlight `alwyas_ff` keyword.
It seemds that the keyword is treated as an indentifier.

Image

prokie:
This should fix #1800

### DIFF
--- a/crates/languageserver/src/server.rs
+++ b/crates/languageserver/src/server.rs
@@ -532,9 +532,7 @@ impl Server {
                     && let VerylSymbolKind::Port(_) = symbol.kind
                 {
                     let token_type = semantic_legend::PROPERTY;
-                    if !is_keyword_token(symbol.token) {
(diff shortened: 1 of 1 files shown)


prokie:
Here is example code on the difference.
This is with current veryl-ls
image
This is with the fix

image

Example code to reproduce issue
```veryl
module jtag_tap2 (
    i_tck : input  clock,
    i_trst: input  reset,
    o_tdo : output logic,
) {

    always_ff {
        if_reset {
            o_tdo = 1'b0;
        }
    }
}
```

dalance (commented):
This is very helpful because I couldn't find the root cause.

I think the keyword check is necessary for `references` only.
This is because the references field may include keyword token.

dalance:
Thanks! I'll merge this PR.

PR #2274, opened 2026-03-02
Fix `unevaluatable_value` error reported for width cast

### LINKED ISSUE #2259: Width cast can't be used in `if_reset` statement
Width cast that the given width is imported from a package can't be used in `if_reset` statement.

Example:

```systemverilog
package Pkg {
    const W: u32 = 8;
}

module ModuleA (
    i_clk: input  clock,
    i_rst: input  reset,
    o_d  : output logic<W>,
) {
    import Pkg::*;

    var d: logic<W>;

    assign o_d = d;

    always_ff {
        if_reset {
            d = 0 as W;
        } else {
            d = d + (1 as W);
        }
    }
}
```

output
```
]8;;https://doc.veryl-lang.org/book/07_appendix/02_semantic_error.html#unevaluatable_value\unevaluatable_value (link)]8;;\

  💥 Value, which is not evaluable at elaboration, can't be used as reset
  │ value
    ╭─[:18:17]
 17 │         if_reset {
 18 │             d = 0 as W;
    ·                 ───┬──
    ·                    ╰── Error location
 19 │         } else {
    ╰────
  help: 
```

taichi-ishitani:
fix veryl-lang/veryl#2259

Currently, IR conversion for cast operation whose RHS is factor defined in other component is treated as error. This causes #2259.

To fix #2259, I implemented the above case.

### DIFF
--- a/crates/analyzer/src/conv/expression.rs
+++ b/crates/analyzer/src/conv/expression.rs
@@ -1,7 +1,7 @@
 use crate::analyzer_error::AnalyzerError;
 use crate::conv::utils::{
-    TypePosition, case_condition, eval_expr, eval_external_symbol, eval_function_call, eval_size,
-    eval_struct_constructor, eval_type, eval_width_select, range_list, switch_condition,
+    TypePosition, case_condition, eval_expr, eval_factor_path, eval_function_call, eval_size,
+    eval_struct_constructor, eval_type, range_list, switch_condition,
 };
 use crate::conv::{Context, Conv};
 use crate::ir::{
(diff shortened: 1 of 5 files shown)


crates/analyzer/src/conv/expression.rs
```diff
@@ -708,76 +728,11 @@ impl Conv<&IdentifierFactor> for ir::Expression {
                 }
             }
         } else {
-            let x = value.expression_identifier.as_ref();
```
taichi-ishitani: The same routine is needed [here](https://github.com/taichi-ishitani/veryl/blob/4aff37f82daecc5cab932cf3a12e50915a9865d7/crates/analyzer/src/conv/expression.rs#L312) too.
Therefore, split this into [`eval_factor_path`](https://github.com/taichi-ishitani/veryl/blob/4aff37f82daecc5cab932cf3a12e50915a9865d7/crates/analyzer/src/conv/utils.rs#L1207).

PR #2277, opened 2026-03-02
Fix formatter panic when vertical align mode is off

prokie:
This is the same fix as was implemented for the emitter in #2251 but for the formatter.


dalance:
I think this PR can be merged, but there seem to be some conflicts.
Could you resolve them?

PR #2291, opened 2026-03-04
Fix/eval type param width

tignear:
In `eval_type`, the `SymbolKind::Parameter` branch only extracts kind from the resolved type, discarding width, array, and signed.  
          
This causes width loss when a type parameter is passed as a generic argument. For example,`param T: type = logic<8>` used in `Inner::<T>` resolves `T` as `logic` (width 1) instead of  `logic<8>`.


dalance:
Thanks!
I'll merge this PR.

Issue #2297, opened 2026-03-06
`unassign_variable` is reported for variable assigned in `for` loop in function


taichi-ishitani:
Veryl compiler reports `unassign_variable` warning for variable assigned in `for` loop in function called before defining it.

Example

```systemverilog
module ModuleA #(
    param N: u32 = 4
) {
    initial {
        func();
    }

    function func() {
        const D: u32 = $clog2(N);
        
        var a: u32;
        for i: u32 in 0..D {
            a = i;
        }
    }
}
```

```console
]8;;https://doc.veryl-lang.org/book/07_appendix/02_semantic_error.html#unassign_variable\unassign_variable (link)]8;;\

  ⚠️ "func.a" is unassigned
    ╭─[:11:13]
 10 │         
 11 │         var a: u32;
    ·             ┬
    ·             ╰── Error location
 12 │         for i: u32 in 0..D {
    ╰────
  help: 
```

PR #2298, opened 2026-03-06
Fix `unassign_variable` warning

taichi-ishitani:
fix veryl-lang/veryl#2297

The caller context should be used to extract the given local function but an empty context is used.
Due to this, a `const` used in the function can't be found and `unassign_variable` is reported becuase `for` loop in the function can't be evaluated.

To fix this issue, the given context should be used if the given function is a local function.

### DIFF
--- a/crates/analyzer/src/conv/context.rs
+++ b/crates/analyzer/src/conv/context.rs
@@ -43,7 +43,6 @@ pub struct Context {
     pub func_paths: HashMap<FuncPath, VarId>,
     pub variables: HashMap<VarId, Variable>,
     pub functions: HashMap<VarId, Function>,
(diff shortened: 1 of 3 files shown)


crates/analyzer/src/conv/context.rs
```diff
@@ -43,7 +43,6 @@ pub struct Context {
     pub func_paths: HashMap<FuncPath, VarId>,
     pub variables: HashMap<VarId, Variable>,
     pub functions: HashMap<VarId, Function>,
-    pub func_call_depth: u32,
```
taichi-ishitani: `func_call_depth` is not used so I removed it.

PR #2302, opened 2026-03-07
Fix formatting issue when const above let

### LINKED ISSUE #1589: format of const above let is wrong
```veryl 
module TopModule {
    const _c   : u32   = 0;
    let _a   : logic = 0;
    let _abcd: logic = 0;
}
```

prokie:
Fix #1589


dalance:
But the current implemantation relies on the char differences between `const` and `var` is 2.
I think adding an align point for variable keyword (`const` `var` `let`) may be better.

For example,

* add something like `align_kind::VAR_KEYWORD` in `crates/aligner/src/lib.rs`.
* add `align_start` and `align_finish` around these keywords like below:

```rust
self.align_start(align_kind::VAR_KEYWORD);
self.r#const.(&arg.r#const);
self.align_finish(align_kind::VAR_KEYWORD);
```

By this change, `const` `var` `let` keyword will be aligned like below:

```
const a: u32 = 0;
var   b: u32;
let   c: u32 = 0;
```

prokie:
Thanks for the comment. I agreem that is a better solutiion compared to mine. I updated the code!

dalance:
Thanks!
I'll merge this PR.

Issue #2262, opened 2026-02-27
Built-in min/max functions


Timmmm:
It would be nice if `min`/`max` worked out of the box. They're extremely common in signal widths, e.g. `var foo: logic<max(N, M)>`. I guess it's a bit tricky because of side effects, e.g. you can't compile it to `M > N ? M : N` because of `max(print("hello"), print("world"))` or whatever?

Issue #2088, opened 2025-12-01
Generic function emitted into the caller namespace


taichi-ishitani:
Currently, only symbols defined within the package or same namespace of a generic component can be used as generic arg.
Therefore, symbols defined within the different namespace of a generic component can't be used as generic arg.
To solve this limitation, I would like to introduce generic functions that are emitted into the caller namespace.

* Such functions are defined within the project namespace but not component namespaces
* Generic components can't be referenced from such functions
    * The above limitation needs to be followed to emit valid SV from generic components

Example

Veryl

```systemverilog
function add::<WIDTH: u32> (
    a: input logic<WIDTH>,
    b: input logic<WIDTH>,
) -> logic<WIDTH> {
    return a + b;
}

module a_module #(
    param WIDTH: u32 = 8,
) (
    i_a: input  logic<WIDTH>,
    i_b: input  logic<WIDTH>,
    o_c: output logic<WIDTH>,
) {
    assign o_c = add::<WIDTH>(i_a, i_b);
}
```

Generated SV

```systemverilog
module a_module #(
    parameter int unsigned WIDTH = 8
)(
    input  var logic [WIDTH-1:0] i_a,
    input  var logic [WIDTH-1:0] i_b,
    output var logic [WIDTH-1:0] o_c
);
    always_comb o_c = mux__WIDTH(i_a, i_b);

    function logic [WIDTH-1:0] add__WIDTH(
        input logic [WIDTH-1:0] a,
        input logic [WIDTH-1:0] b
    );
        return a + b;
    endfunction
endmodule
```

PR #2332, opened 2026-03-14
added min and max

prokie:
Added min and max function to solve #2262. I think there needs to be some check to make sure that the args is not string or a function, but unsure how to implement that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -120,9 +120,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
-version = "1.0.13"
+version = "1.0.14"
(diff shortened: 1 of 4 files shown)


taichi-ishitani:
I think it is prefer that these functions are implemented as Veryl functions and provided as parts of the `std` library but not system functions.
However, due to limitation of generic args, such functions can't be implemented. To resolve the limitation, #2088 is needed.

prokie:
Okay, I understand. I guess we can close this pull request then.

taichi-ishitani:
#2088 has been implemented.

Issue #2349, opened 2026-03-17
Accept expression and factor type as generic args


taichi-ishitani:
Currently, expression and factor type cannot be used as generic args.
Therefore, syntax error is reported for like following code.

```systemverilog
module ModuleA::<A: u32, B: u32, C: u32> {
    inst u: ModuleB::<A + B, logic<C>>;
}
```

To resolve this kind of inconvenience, expression and factor type should be accepeted as generc args.

taichi-ishitani:
`>>` in the above example may be treated as a shift operator.
To resolve this confilicat, special notation is needed for factor type given as generic arg.

taichi-ishitani:
Another solution.
Introduce a new variable type which can be given as generc arg.

```systemverilog
module ModuleA::<A: u32, B: u32, C: u32> {
    genvar A_B   : u32  = A + B;
    genvar C_TYPE: type = logic<C>:
    inst u: ModuleB::<A_B, C_TYPE>;
}
```

Prefix `genvar` is the same with SV's keyword `genvar`. I consern that it may be confusable.

taichi-ishitani:
`A_B` and `C_TYPE` act as constant so `genvar` is not suitable.
Is `genconst` is better?

omasanori:
> `>>` in the above example may be treated as a shift operator.
> To resolve this confilicat, special notation is needed for factor type given as generic arg.
> 

Haven't C++ solved a similar problem a decade ago?

dalance:
> Haven't C++ solved a similar problem a decade ago?

Of cource, we can allow it like C++, but it causes parser's complexity.
Unless there's a strong motivation to use shift/compare operators within generic arguments,
I think we should avoid introducing unnecessary complexity.

taichi-ishitani:
For generic module, `genconst` may be acceptable.

```systemverilog
genconst A_B   : u32 = A + B;
genconst C_TYPE: type = logic<C>;
inst u: ModuleA::<A_B, C>;
```

However, for generic function, defining `genconst` may be little complicated.

```systemverilog
genconst A_B   : u32 = A + B;
genconst C_TYPE: type = logic<C>;
always_comb {
    a = func::<A_B, C_TYPE>(b, c);
}
```

I think it is more preferred to introduce some solutions for the above confliction and allow to put expression to generic arg list directly.

```systemverilog
always_comb {
    a = func::<A + B, type(logic<C>)>(b, c); // introduce `type()` notation
[...]

PR #2412, opened 2026-03-27
Itroduce `gen` declaration

taichi-ishitani:
close veryl-lang/veryl#2349

`gen` declaration is a new declaration like `const` declaration for `generics` feature. This is used to define `genreic` const which can be given as generic asg.

By using `gen` declaration, you can provide result of expression and `variable` type declaratoin to  generic arg like below.

```systemverilog
module ModuleA::<W: u32, T: type) (
    a: output logic<W>,
    b: output T       ,
) {}
module ModuleB::<A: u32, B: u32, C: u32, D: u32> {
    gen W: u32  = A + B;
    gen T: type = logic<C, D>;
    inst u: ModuleA::<W, T> (a: _, b: _);
}
```

Expression in `gen` declaration has the same limitation with generic arg becuase `generic` consts are used as gneric arg. Therefore, the given expression can include following factors:

* Literal
* Symbol defined in package
* Genreic parameter
* Generic const


crates/analyzer/src/conv/utils.rs
```diff
@@ -1519,17 +1547,30 @@ pub fn eval_external_symbol(
...
                     let (mut x, _) = eval_expr(context, None, x, false)?;
                     x.token = token;
-                    return Ok(ir::Factor::Value(x));
+                    Some(ir::Factor::Value(x))
                 }
-                EnumMemberValue::UnevaluableValue => (),
+                EnumMemberValue::UnevaluableValue => None,
+            };
+
+            if let Some(mut factor) = factor {
```
taichi-ishitani: Type kind of enum member should be its parent enum. The replacement below is for this.

PR #2442, opened 2026-04-01
Fix array literal default expansion generating extra element

tignear:
In eval_array_literal, the remaining count for default fill was computed as `target_len - (x.len() - 1)` which equals `target_len` when `x.len() == 1`, producing one too many default elements. Changed to `target_len - len` where `len` tracks the number of explicitly provided elements.


dalance:
That's certainly true. Thanks!

PR #2447, opened 2026-04-02
Add `zero_extend`, `one_extend`, `sign_extend` and `truncate` as std utility functions

### LINKED ISSUE #2266: Explicit zero/sign/one extension and truncation functions.
SystemVerilog does implicit zero extension and truncation, which is really really bad. It seems like Veryl doesn't really attempt to detect that, but underlying tools do (e.g. Verilator). For example if I do

```
var a: logic<8>;
var b: logic<16>;
always_comb {
  b = a;
}
```

Verilator will give an error by default (which is good).

It would be nice if Veryl had explicit syntax to say "no I really do want to extend/truncate this". I would suggest this (these are the same functions as in [Sail](https://github.com/rems-project/sail):

* `zero_extend(..)`
* `sign_extend(..)`
* `one_extend(..)` - not actually a built-in in Sail but it is useful.
* `truncate(..)`

This may be quite tricky to compile to SV... but maybe something like this would work?

* `b <= {{($bits(b)-$bits(a)){1'b0}}, a};`
* `b <= {{($bits(b)-$bits(a)){1'b1}}, a};`
* `b <= {{($bits(b)-$bits(a)){a[$bits(a)-1]}}, a};`
* `b <= a[$bits(b)-1:0];`

Also they become context sensitive so maybe they'd need to only be allowed in certain locations. Sail has a super sophisticated type system to make it work.

Anyway just an idea.

taichi-ishitani:
close veryl-lang/veryl#2266

### DIFF
--- a/crates/std/veryl/src/utility_functions/extend.veryl
+++ b/crates/std/veryl/src/utility_functions/extend.veryl
@@ -1,0 +1,74 @@
+pub function zero_extend::<FROM_TYPE: type, TO_TYPE: type,> (
+    from: input FROM_TYPE,
+) -> TO_TYPE {
+    const FROM_WIDTH  : p32 = $bits(FROM_TYPE);
+    const TO_WIDTH    : p32 = $bits(TO_TYPE);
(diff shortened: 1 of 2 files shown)


dalance (approved):

Timmmm:
Very nice, thanks!

PR #2451, opened 2026-04-02
Add `select_*` and `dispatch_*` functions

taichi-ishitani:
This PR is to extract functions that implement mux/demux functionality outside of the `mux`/`demux` modules.
These functions are named `select_*` and `dispatch_*` because `mux` and `demux` has already been used.


dalance:
I think the implementation is no problem, but the naming may be improved.
I feel `distribute` shows "spread something evenly" like `{data repeat N}`.
There are some ideas, how about them?

* `inject` / `insert`: inserting data to a specific position of array
* `dispatch`: routing data to a specific position of array

taichi-ishitani:
Thank you for your feedback.
I renamed these functions with `dispatch_*`.

PR #2498, opened 2026-04-11
Support `for` loops with dynamically determined ranges

taichi-ishitani:
fix veryl-lang/veryl#2490

Currently, simulator does not support `for` loop with dynamically determined range.
Due to this, `$std::select_vector` and `$std::select_one hot` are not simulated correctlly.

This PR is to add such `for` loop support veryl simulator.

### DIFF
--- a/crates/analyzer/src/conv/statement.rs
+++ b/crates/analyzer/src/conv/statement.rs
@@ -631,16 +631,15 @@ impl Conv<&ForStatement> for ir::StatementBlock {
             .for_statement_opt0
             .as_ref()
             .map(|x| (x.assignment_operator.as_ref(), x.expression.as_ref()));
+        let for_range = build_for_range(context, &value.range, rev, step)?;
 
-        if context.in_sequential_block {
-            let for_range = build_for_range(context, &value.range, rev, step)?;
+        if context.in_sequential_block || for_range.is_dynamic() {
             return build_for_statement(context, value, &r#type, clock_domain, for_range, token);
         }
 
-        let range = eval_for_range(context, &value.range, rev, step, token)?;
+        let range = eval_for_range(context, &for_range, token)?;
 
         let mut ret = ir::StatementBlock::default();
-
         for i in range {
             let label = format!("[{}]", i);
             let label = resource_table::insert_str(&label);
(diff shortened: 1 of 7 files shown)


dalance:
Generally, Claude Code tends to generate quite verbose comments, so I think it's better to have it revise them to keep them to more simple.

dalance:
I considered how to handle Claude-generated code.
I don't think there's a problem with adding the generated code itself, but I think there might be copyright issues if the commit signature includes Claude Code.
(Normally, copyright arises when code is written, but AI-generated code is not considered to have copyright,
so it may not be possible to license it properly.)
Therefore, instead of letting Claude directly generate commits, please carefully review the content and commit it yourself.

crates/analyzer/src/conv/statement.rs
```diff
@@ -631,16 +631,15 @@ impl Conv<&ForStatement> for ir::StatementBlock {
             .for_statement_opt0
             .as_ref()
             .map(|x| (x.assignment_operator.as_ref(), x.expression.as_ref()));
+        let for_range = build_for_range(context, &value.range, rev, step)?;
 
-        if context.in_sequential_block {
-            let for_range = build_for_range(context, &value.range, rev, step)?;
+        if context.in_sequential_block || for_range.is_dynamic() {
```
taichi-ishitani, line 636: Avoid unrolling `for` loop with dynamically determined range because its range can't be evaluated at this time.

crates/simulator/src/ir/statement.rs
```diff
@@ -673,33 +670,302 @@ pub struct CompiledBlockStatement {
...
+                    signed: *signed,
+                }
+            }
+        }
+    }
+
+    pub fn get_value(&self) -> u64 {
+        match self {
+            Self::Const(x) => *x,
+            Self::Variable {
```
taichi-ishitani: For loop bound is evaluated dynamically when it is given by variable.

crates/simulator/src/ir/statement.rs
```diff
@@ -685,33 +682,325 @@ pub struct CompiledBlockStatement {
...
+                Self::Expression(Box::new(exp))
+            }
+        }
+    }
+
+    pub fn eval(&self, mask_cache: &mut MaskCache) -> u64 {
+        match self {
+            Self::Const(x) => *x,
+            Self::Expression(exp) => {
+                let val = exp.eval(mask_cache);
```
taichi-ishitani, line 753: For loop bound is evaluated dynamically if it is given by variable.

taichi-ishitani:
I discussed it with @dalance .
He is considering changing how to handle `for` loop in IR context; handle `for` loop as is.

So I close this PR.

PR #2538, opened 2026-04-20
Add function calculating leading zero count

taichi-ishitani:
This PR adds a function calculating leading zero count as `std` library.
Implementation is based on the paper below.
https://zenodo.org/records/14810176

For now, test `test_lzc_128` is commented out because simulator panics whem simulating this test case.

```console
thread '<unnamed>' (57563) panicked at /home/taichi/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cranelift-codegen-0.130.1/src/isa/x64/lower/isle.rs:76:5:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

thread 'main' (57554) panicked at crates/veryl/src/cmd_test.rs:238:54:
called `Result::unwrap()` on an `Err` value: Any { .. }
```

### DIFF
--- a/crates/std/veryl/src/lzc/lzc_common_pkg.veryl
+++ b/crates/std/veryl/src/lzc/lzc_common_pkg.veryl
@@ -1,0 +1,95 @@
+pub package lzc_common_pkg {
+    struct result::<N: p32> {
+        all_zero: logic           ,
+        count   : logic<$clog2(N)>,
+    }
+
+    function lzc4 (
+        bits: input logic<4>,
+    ) -> result::<4> {
+        var ret         : result::<4>;
+        ret.all_zero = bits == '0;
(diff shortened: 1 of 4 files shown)


dalance:
I'm investigating the simulator panic.

dalance:
#2544 fixed the panic.

dalance:
Additioally, `for i: u32 in N_SPLITS ` is wrong loop, so simulation will be failed.
Please use `for i: u32 in 0..N_SPLITS`.

taichi-ishitani:
> Additioally, `for i: u32 in N_SPLITS ` is wrong loop, so simulation will be failed. Please use `for i: u32 in 0..N_SPLITS`.

Thank you for your update.
I fixed the implementation and confirmted all tests for lzc were passed.

PR #2577, opened 2026-04-24
Add `$std::utils::clog2_clipped`

### LINKED ISSUE #2571: Add clipped version of `$clog2`
Ceiling result of `$clog2` to 1 like below is commonly used to calculate bit width.

```systemverilog
const WIDTH: u32 = if N >: 1 ? $clog2(N) : 1;
```

Add clipped version of `$clog2` as std library for this usage.

taichi-ishitani:
close veryl-lang/veryl#2571

### DIFF
--- a/crates/std/veryl/src/pkg/utils.veryl
+++ b/crates/std/veryl/src/pkg/utils.veryl
@@ -1,0 +1,32 @@
+pub package utils {
+    function clog2_clipped (
+        x  : input u32,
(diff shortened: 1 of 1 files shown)


dalance:
About function name `$clog2`, I feel the meaning of the second arg is not clear.
If it is `$clog2_clipped`, I can guess that the second is value to use for clipping.
How about it?

PR #2635, opened 2026-05-18
Add profiler for `analyzer_pass2`

taichi-ishitani:
Converting AST to IR is one of main routines in `analyze` phase. Profiler for `analyze_pass2` is added to analyze performance of IR conversion.
It collects elapsed time to convert AST to IR for each compoenent declarations and output them to TOML file.

### DIFF
--- a/crates/analyzer/src/conv/conv_profiler.rs
+++ b/crates/analyzer/src/conv/conv_profiler.rs
@@ -1,0 +1,74 @@
+use crate::HashMap;
+use miette::{self, IntoDiagnostic, Result, WrapErr};
+use serde::Serialize;
+use std::fs;
+use std::sync::{Arc, Mutex};
+use std::time::{Duration, Instant};
+
+#[derive(Clone, Default, Serialize)]
+pub struct ConvStats {
+    pub count: u64,
+    pub total_us: u64,
+    pub avg_us: u64,
+    pub min_us: u64,
+    pub max_us: u64,
+}
+
+#[derive(Clone, Default)]
+pub struct ConvProfile {
+    pub entries: HashMap<String, ConvStats>,
+}
+
+impl ConvProfile {
+    pub fn record(&mut self, name: &str, duration: Duration) {
+        let us = duration.as_micros() as u64;
+        let entry = self.entries.entry(name.to_string()).or_default();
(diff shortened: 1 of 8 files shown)


crates/analyzer/src/conv/ir.rs
```diff
@@ -125,6 +125,8 @@ impl Conv<&ModuleDeclaration> for ir::Module {
         let mut context = Context::default();
         context.inherit(upper_context);
 
+        let _guard = context.conv_profile_guard(value.identifier.text().to_string());
```
dalance: Strictly speaking, this TLS access and memory allocation via `to_string` is performed every time, even when profiler is disabled, so it's not zero cost. What do you think?
It might be better to pass `to_string` as a closure (the closure won't be evaluated if the profiler is `None`) or to make the profile a feature and enable it at compile time.

taichi-ishitani: I changed type of the argument to `StrId` and removed `to_string` func call.

crates/analyzer/src/conv/conv_profiler.rs
```diff
@@ -0,0 +1,71 @@
+use crate::HashMap;
+use serde::Serialize;
+use std::sync::{Arc, Mutex};
+use std::time::{Duration, Instant};
+
+pub trait ConvProfiler {
```
dalance: Why is this trait? Do you intend to expand it for another profiler?
If so, more generic name may be suitable.
If not, `record` can be added to `ConvProfile` directly.

crates/veryl/src/cmd_build.rs
```diff
@@ -115,6 +115,9 @@ impl CmdBuild {
                 .defines
                 .insert(resource_table::insert_str(name));
         }
+        if std::env::var("VERYL_CONV_PROFILE").is_ok() {
```
dalance: Checking `VERYL_CONV_PROFILE` is duplicated through some lines/files.
How about moving it to `enable_conv_profiler` and something like `finalize_conv_profiler`?
Probably, the check and extract `path` are needed only once.

taichi-ishitani:
@dalance ,

I fixed feedbacks from you.
Could you please reveiw again?

PR #2730, opened 2026-06-04
Add --out-dir option to veryl build for redirecting output location

### LINKED ISSUE #2685: [Feature] Add a standalone compile command with support for specifying the output location (similar to `cc`)
## Background

I am currently trying to integrate veryl with other build systems (such as xmake or build.rs). I encountered a requirement to output the results to a specified directory. Currently, `veryl build` can only obtain the output path from the configuration file. 

## Possible Solution

Provide a compile command similar to `cc`, which processes only one file at a time and allows specifying the output. 

```bash
verylc path/to/source.veryl -o path/to/dist.vs --clock_type=posedge --reset_type=async_low
# or
verylc path/to/source.veryl -o path/to/dist.vs --config=Veryl.toml
```

By default, it should not follow any project configuration. All compilation options should be passed via command-line arguments, or a project configuration file can be explicitly specified. 

In this case, the invoker should be responsible for handling the compilation configuration and ensuring consistency.

## Another possible solution

Add an option to `veryl build`, such as `--base-dir=path/to/base`. 

In this case, all configs like

```toml
target = {type = "directory", path = "dist"}
sourcemap_target = {type = "directory", path = "map"}
```

will be written to `path/to/base/dist` and `path/to/base/map`.

mvanhorn:
## Summary
`veryl build --base-dir <DIR>` redirects build outputs (generated `.sv`, `.sv.map`) to a caller-specified directory while sources keep resolving against the project path. Without the flag, behavior is byte-for-byte unchanged.

## Why this matters
#2685 asks for a way to direct `veryl build` output when integrating with external build systems (xmake, build.rs), and dalance steered toward this exact option: "The second option (`--base-dir`) is probably easier to consider. This should integrate seamlessly into the current structure."

Implementation follows that seamless-integration framing:
- `OptBuild` gains an optional `--base-dir <DIR>` (`crates/veryl/src/lib.rs`); `cmd_build` resolves relative paths against the current working directory and stores the result as `Metadata::output_base_override` (a `#[serde(skip)]` field, never read from Veryl.toml)
- `Metadata::paths` computes destinations against the override when present: `Target::Directory` and `Target::Bundle` swap their project-path base for the override, and `Target::Source` (normally emits next to each source) preserves the source-relative layout under the override; sourcemap paths follow the same rule
- `cmd_test` passes `base_dir: None`, so `veryl test` is untouched

## Testing
Verified on a fresh `veryl new` project (target `{type = "directory", path = "target"}`) with the built binary:
- no flag: `target/hello.sv` exactly as before
- `--base-dir /abs/path`: emits `/abs/path/target/hello.sv` + `.sv.map`; project tree gets no new outputs
- `--base-dir relout` (relative): emits `<cwd>/relout/target/hello.sv`
- the generated filelist references the redirected destinations
- `cargo check -p veryl -p veryl-metadata`, `cargo fmt --check`, and `cargo test -p veryl-metadata` pass

The filelist itself still lands at the project path; happy to redirect it too if you'd prefer everything out-of-tree.

Closes #2685

### DIFF
--- a/crates/metadata/src/metadata.rs
+++ b/crates/metadata/src/metadata.rs
@@ -66,6 +66,11 @@ pub struct Metadata {
     pub lockfile: Lockfile,
     #[serde(skip)]
     pub build_info: BuildInfo,
+    /// Output base directory override (e.g. `veryl build --base-dir`).
+    /// When set, build outputs are emitted relative to this directory
+    /// instead of the project path. Never read from Veryl.toml.
+    #[serde(skip)]
+    pub output_base_override: Option<PathBuf>,
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize, Hash, PartialEq, Eq, PartialOrd, Ord)]
@@ -310,6 +315,12 @@ impl Metadata {
         };
 
         let base = self.project_path();
+        // Build outputs may be redirected (e.g. `veryl build --base-dir`);
+        // sources are always resolved against the project path.
+        let out_base = self
+            .output_base_override
+            .clone()
+            .unwrap_or_else(|| base.clone());
         let mut ret = Vec::new();
 
         // Pre-canonicalize explicit file args once so we can route each to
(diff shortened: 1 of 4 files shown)


dalance:
The changes seem fine, but I'd like to wait to merge until I can confirm whether the `--base-dir` option properly resolves the original issue.

dalance:
I tried to specify the absolute path outside the project, and it failed.

```
$ veryl build --base-dir /tmp
Error:   × prefix not found 
```

PFiS1737:
If possible, I would also like the `*.f` file and the `dependencies` directory to be generated within this directory.

Additionally, would naming this option `--out-dir` or `--build-dir` be more appropriate?

mvanhorn:
Updated in ee66aa9:

@dalance the absolute-path case is fixed - the "prefix not found" came from assuming the output dir was inside the project root. Verified:

```
$ veryl build --out-dir /tmp/outside-out   # from a scratch project
[INFO] Output filelist (/tmp/outside-out/smoke_prj.f)
```

@PFiS1737 both requests are in: the `*.f` filelist and the `dependencies` directory are now generated inside the specified directory (along with the generated .sv under `target/`), and the option is renamed to `--out-dir`. Happy to switch to `--build-dir` instead if that reads better.

Default behavior without the option is unchanged. Relative paths resolve against the current working directory.

dalance:
Looks good!

mvanhorn:
Thanks for the merge, @dalance. The --out-dir option makes veryl build fit cleanly into external build trees.

PR #2753, opened 2026-06-06
Pass verilator args to cocotb runner

PoppyClick:
This PR applies [test.verilator] arguments to cocotb  `runner.build(build_args)` and `runner.test(test_args)` parameters.

A better solution would be to create a [test.cocotb] section, as verilator might not be the only backend in the future.

I've tested it by providing two verilog files to compile_args and nothing to simulate_args:

```toml
[test.verilator]
# Replace "~/home/{user}/Tools/oss-cad-suite/share/yosys" with the output of "yosys-config --datdir" command
compile_args = ["/home/{user}/Tools/oss-cad-suite/share/yosys/gatemate/cells_sim.v",
                "/home/{user}/Tools/oss-cad-suite/share/yosys/gatemate/cells_bb.v"]
```

### DIFF
--- a/crates/veryl/src/runner/cocotb.rs
+++ b/crates/veryl/src/runner/cocotb.rs
@@ -1,4 +1,5 @@
 use crate::runner::{Runner, copy_wave};
+//use fern::meta;
 use futures::prelude::*;
 use log::{error, info};
 use miette::{IntoDiagnostic, Result, WrapErr};
--- a/crates/veryl/src/runner/cocotb.rs
+++ b/crates/veryl/src/runner/cocotb.rs
@@ -169,7 +169,7 @@ impl Runner for Cocotb {
         let module = format!("{}_{}", metadata.project.name, top.unwrap());
 
         let (py_waves, build_args) = match wave.then_some(metadata.test.waveform_format) {
-            Some(WaveFormFormat::Vcd) => ("True", "['--trace']"),
+            Some(WaveFormFormat::Vcd) => ("True", "['--trace-vcd']"),
             Some(WaveFormFormat::Fst) => ("True", "['--trace-fst', '--trace-structs']"),
             None => ("False", "[]"),
         };
--- a/crates/veryl/src/runner/cocotb.rs
+++ b/crates/veryl/src/runner/cocotb.rs
@@ -168,7 +168,7 @@ impl Runner for Cocotb {
         sources = format!("[{}]", sources.strip_suffix(',').unwrap());
 
         let module = format!("{}_{}", metadata.project.name, top.unwrap());
-    
+
         // Create storage for build and test arguments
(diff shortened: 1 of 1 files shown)


crates/veryl/src/runner/cocotb.rs
```diff
@@ -168,11 +169,53 @@ impl Runner for Cocotb {
...
+        let mut build_args = String::new();
+        let mut test_args = String::new();
+
+        if !vec_build_args.is_empty() {
+            for arg in &vec_build_args {
+                build_args.push_str(&format!("'{arg}',"));
+            }
+            // If wave_args isn't empty, add it to the build args.
+            if !wave_args.is_empty() {
+                build_args.push_str(wave_args);
```
dalance, line 203: If `vec_build_args` is empty, adding `wave_args` will be skipped.
This probably break wave dump.

crates/veryl/src/runner/cocotb.rs
```diff
@@ -168,11 +169,53 @@ impl Runner for Cocotb {
...
+            build_args.push_str("[]");
+        }
+
+        if !vec_test_args.is_empty() {
+            for arg in &vec_test_args {
+                test_args.push_str(&format!("'{arg}',"));
+            }
+            test_args = format!("[{}]", test_args.strip_suffix(',').unwrap());
+        } else {
+            test_args.push_str("[]");
```
dalance, line 216: Adding `[]` and `strip_prefix(',')` are used repeatedly.
So these can be extracted to a function like `format_py_array(array: &[String]) -> String`.

crates/veryl/src/runner/cocotb.rs
```diff
@@ -168,11 +169,53 @@ impl Runner for Cocotb {
...
-            Some(WaveFormFormat::Vcd) => ("True", "['--trace']"),
-            Some(WaveFormFormat::Fst) => ("True", "['--trace-fst', '--trace-structs']"),
-            None => ("False", "[]"),
+        // Create storage for build and test arguments
+        let mut vec_build_args = Vec::new();
+        let mut vec_test_args = Vec::new();
+
+        // Parse wave arguments
+        let (py_waves, wave_args) = match wave.then_some(metadata.test.waveform_format) {
+            Some(WaveFormFormat::Vcd) => ("True", "'--trace-vcd',"),
```
dalance, line 178: I'm not very familiar with Verilator options, but is `--trace-vcd` better than `--trace`?

dalance:
I want to release this feature in the next version, but I hadn't received a response for some time.
Therefore, since #2761 addressed my points, I've decided to merge that instead.

PoppyClick:
Yes sorry, I failed to take the time to make the changes that you kindly pointed (thank you for that :) ). #2761 covers the same point without any of the issues of my PR, and it looks like they made better code for it.

PR #2761, opened 2026-06-08
Cocotb runner inherits the Verilator build and simulate args

jbeaurivage:
Currently there is no way to add build/simulation arguments to the Verilator backend when using cocotb runners. This PR makes the cocotb runner forward the build and simulation arguments specified in the `[test.verilator]` section of `Veryl.toml`.


jbeaurivage:
Whoops, just realized this is a dupe of #2753. Feel free to merge whichever you prefer, or close this PR if you want.

dalance:
Thanks!
I'll review and merge #2753 instead of this PR because I've already started check it.

dalance:
Since I haven't received a reply from the author of #2753, I'd like to merge this one.

PR #2851, opened 2026-06-17
fix(analyzer): report an error for a bare for-loop range

### LINKED ISSUE #2543: `for i in N` is accepted, but incorrect
`for i in N` is valid syntax, but semantically incorrect.
It should be reported as error.

mvanhorn:
## Summary

Reports an analyzer error when a for-loop range is a bare expression with no range operator (`for i in N`).

The grammar defines `ForStatement: For Identifier In [Rev] Range ...` where `Range: Expression [RangeOperator Expression]`, so the parser accepts a bare expression as a "range". Semantically this has no valid loop meaning, but it was previously accepted silently. This change adds a new `invalid_for_range` analyzer error that fires when a for-loop's `Range` has no `..`/`..=` operator, for both the procedural for-statement and the `for ... :blk` generate-for declaration.

The separately-discussed idea of treating `for i in N` as sugar for `0..N` is intentionally not implemented here; this PR only rejects the bare form, which is the issue's standing ask.

## Why this matters

`for i in N` looks like a count-based loop but is parsed as a single-scalar "range" that cannot describe an iteration, so it should be a compile-time error rather than silently-accepted, meaningless code. See #2543.

## Testing

- `cargo test -p veryl-analyzer` (345 passing, including the new `invalid_for_range` test)
- Full workspace test suite green (663 tests across parser, analyzer, emitter, and integration)
- `cargo fmt --check`
- `cargo clippy -p veryl-analyzer -- -D warnings`

The new test covers the error path for both the statement form (`for i in 4 { }`) and the generate-for form (`for i in 4 :blk { }`), plus the happy path for `0..4` and `0..=4` ranges.

Fixes #2543

### DIFF
--- a/crates/analyzer/src/analyzer_error.rs
+++ b/crates/analyzer/src/analyzer_error.rs
@@ -1997,6 +1997,7 @@ impl AnalyzerError {
             AnalyzerError::InvalidEmbedIdentifier { input, .. } => input,
             AnalyzerError::InvalidEnumVariant { input, .. } => input,
             AnalyzerError::InvalidFactor { input, .. } => input,
+            AnalyzerError::InvalidForRange { input, .. } => input,
             AnalyzerError::InvalidForStep { input, .. } => input,
             AnalyzerError::InvalidIdentifier { input, .. } => input,
             AnalyzerError::InvalidImport { input, .. } => input,
--- a/crates/analyzer/src/analyzer_error.rs
+++ b/crates/analyzer/src/analyzer_error.rs
@@ -709,6 +709,21 @@ pub enum AnalyzerError {
 
     #[diagnostic(
         severity(Error),
+        code(invalid_for_range),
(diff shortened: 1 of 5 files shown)


dalance:
Looks good!
Could you add snapshot test for new error `InvalidForRange`?

* put `testcases/error/invalid_for_range.veryl` including InvalidForRange error.
* `cargo test -p veryl-tests invalid_for_range`, this command creates a new snapshot and fail.
* check `crates/tests/src/snapshots/invalid_for_range.snap.new`, the message and error location is correct or not.
* If ok, cp `crates/tests/src/snapshots/invalid_for_range.snap.new` to `crates/tests/src/snapshots/invalid_for_range.snap`
* `cargo test -p veryl-tests invalid_for_range`, this command checks error message is equal to the above `.snap`, and success.
* commit `testcases/error/invalid_for_range.veryl` and `crates/tests/src/snapshots/invalid_for_range.snap`

PR #2934, opened 2026-06-30
Expose resolved dependency metadata for external tools

### LINKED ISSUE #2648: Expose resolved dependency and cache metadata for external integration tools
Hi, thanks for working on Veryl.

We are evaluating Veryl as a frontend tool for large commercial SoC development, including in-house IP development, third-party commercial IP integration, dependency/version management, and downstream EDA flow integration.

After thinking more about our filelist and EDA flow requirements, I agree that many of them may not be appropriate to integrate directly into Veryl itself. They may be better handled by an external tool layered on top of Veryl.

However, for such a tool to work well, it would need a stable way to reuse Veryl's dependency resolution result and dependency cache, instead of reimplementing a parallel dependency resolver or cloning all dependencies again.

## Background

In a large SoC project, we may have many Veryl projects and third-party commercial SystemVerilog IPs. Some dependency repositories may contain both:

- Veryl source code
- vendor-delivered SystemVerilog source code and filelists

For commercial third-party IP, translating the vendor SystemVerilog into Veryl is usually not the preferred flow. Although `veryl translate` may be useful for migration in some cases, converting vendor-delivered RTL can change the ownership, review, verification, support, or liability boundary of the IP. In those cases, we would prefer to keep the original vendor SystemVerilog source and filelist as the source of truth.

At the same time, we still want these SV resources to live in the same repositories as the related Veryl code, so the top-level SoC can use Veryl's dependency/version management to select consistent versions of all frontend resources.

## Why an external tool may be better

There are many downstream filelist requirements that can be project-, company-, or EDA-tool-specific, for example:

- selecting different technology-library files for generic simulation, ASIC simulation, synthesis, emulation, FPGA, or GLS;
- choosing between simulation-only SRAM/clock-gate models, foundry models, synthesizable FPGA/emulation models, and netlist models;
- preserving or flattening nested `-f` filelists;
- emitting `+define+` options inline or into a separate define file;
- adapting filelist syntax for different commercial EDA tools.

These policies can become quite flow-specific. So instead of asking Veryl to implement all of this filelist logic, a separate tool could generate final EDA filelists from project-specific metadata.

But that external tool needs to know the full Veryl dependency graph and where each resolved dependency is available locally.

## What the external tool would need from Veryl

Ideally, Veryl could provide a stable interface to query resolved dependency information, such as:

- dependency name / namespace used by the parent project;
- original project name;
- source repository or local path;
- resolved version and revision for repository dependencies;
- dependency edges, including transitive dependencies;
- whether a dependency came from `git`, `github`, or `path`;
- the local filesystem path where the dependency is available after `veryl build` or `veryl update`;
- enough identity information to distinguish multiple versions of the same project in one resolved graph.

With that information, an external tool could read additional project-owned metadata, such as SV filelist manifests, technology-library selection rules, or EDA-specific output policies, without needing Veryl itself to understand those policies.

It would also be helpful if `Veryl.toml` had a documented extension mechanism for external tools. For example, an external filelist tool should not need to maintain a separate project-level TOML file just to describe SV filelists or flow-specific metadata. Ideally, tool-specific metadata could live in `Veryl.toml` under a namespace that Veryl reserves for external tools, so it will not conflict with any future Veryl built-in keys.

For example, this could be something like:

```toml
[tool.my_filelist_tool]
sv_filelists = ["filelists/rtl.f"]
```

The exact namespace is not important. The important part is that external tools have a stable place to store their own metadata in `Veryl.toml` without risking future incompatibility with Veryl's own manifest schema.

## Current understanding

From a quick look at the current implementation, `Veryl.lock` appears to already contain much of the resolved dependency graph: project entries, sources, versions, revisions, and dependency edges.

Veryl also already populates a dependency cache during normal dependency resolution/build. From the implementation, the cache path seems to be derived from Veryl's OS cache directory, and repository dependencies are stored under a generated cache slot.

However, I am not sure whether either of these should be treated as a stable external interface:

- Is the `Veryl.lock` schema intended to be parsed by third-party tools?
- Is the cache directory layout intended to be stable?
- Is there a supported command/API to map each lockfile dependency entry to its local cached source path?

I noticed that `veryl metadata --format json` exists, but I am not sure whether it is intended to expose the resolved lockfile graph and local dependency paths for external tooling.

## Possible interface

One possible direction could be extending an existing metadata command or adding a dedicated dependency export command, for example:

```console
veryl metadata --format json --include-resolved-dependencies
```

or:

```console
veryl deps --format json
```

The exact command name and JSON schema are not important to me. The important part is having a documented, stable interface that external tools can rely on.

For example, the output could include entries like:

```json
{
  "dependencies": [
    {
      "name": "pcie_ss",
      "project": "pcie_ss",
      "source": {
        "type": "git",
        "url": "https://example.com/pcie_ss.git",
        "version": "1.2.3",
        "revision": "..."
      },
      "local_path": "...",
      "dependencies": ["..."]
    }
  ]
}
```

Then external tools could reuse Veryl's dependency resolution and cache, while keeping flow-specific filelist generation outside Veryl.

## Questions

Before I spend time designing or implementing external tooling around this, I would like to ask:

1. Is `Veryl.lock` intended to be a stable input for third-party tools, or should external tools avoid parsing it directly?
2. Is there currently a supported way to get the local cache path for each resolved dependency?
3. Would it be reasonable for Veryl to expose a stable resolved-dependency metadata interface for external tools?
4. Is there, or could there be, a reserved namespace in `Veryl.toml` for external tool metadata, so tools can avoid maintaining separate config files?
5. If yes, would maintainers prefer extending `veryl metadata`, adding a new command, or some other approach?

If this direction sounds reasonable, I can prepare a more concrete proposal for the metadata schema and how an external filelist/IP integration tool would consume it.

pxlsy2000:
## Summary
- add a versioned metadata JSON output shape for resolved project graphs
- include root and dependency metadata, source identity, and local paths for external tools
- reserve and preserve external tool metadata from `Veryl.toml`

## Testing
- `cargo fmt --all -- --check`
- `cargo test -p veryl-metadata`
- `cargo test -p veryl cmd_metadata`

Closes #2648

### DIFF
--- a/crates/metadata/src/lockfile.rs
+++ b/crates/metadata/src/lockfile.rs
@@ -93,35 +93,15 @@ impl LockSource {
 #[serde(deny_unknown_fields)]
 pub struct LockSourceRepository {
     uuid: Uuid,
-    url: UrlPath,
-    path: PathBuf,
-    project: String,
-    version: Version,
-    revision: String,
+    pub url: UrlPath,
+    pub path: PathBuf,
+    pub project: String,
+    pub version: Version,
+    pub revision: String,
     r#override: Option<PathBuf>,
 }
 
 impl LockSourceRepository {
-    pub fn url(&self) -> &UrlPath {
-        &self.url
-    }
-
-    pub fn path(&self) -> &Path {
-        &self.path
-    }
-
-    pub fn project(&self) -> &str {
-        &self.project
-    }
-
-    pub fn version(&self) -> &Version {
-        &self.version
-    }
-
-    pub fn revision(&self) -> &str {
-        &self.revision
-    }
-
     pub fn local_path(&self) -> Result<PathBuf, MetadataError> {
         Ok(Lockfile::dependency_path(&self.url, &self.path, &self.revision)?.join(&self.path))
     }
--- a/crates/metadata/src/lockfile.rs
+++ b/crates/metadata/src/lockfile.rs
@@ -73,6 +73,20 @@ impl LockSource {
             LockSource::Path(_) => None,
         }
     }
+
+    pub fn project(&self) -> Option<&str> {
+        match self {
+            LockSource::Repository(x) => Some(&x.project),
+            LockSource::Path(_) => None,
+        }
+    }
+
+    pub fn local_path(&self, root_path: &Path) -> Result<PathBuf, MetadataError> {
+        match self {
+            LockSource::Repository(x) => x.local_path(),
+            LockSource::Path(path) => Ok(root_path.join(path)),
+        }
+    }
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
@@ -85,6 +99,32 @@ pub struct LockSourceRepository {
     version: Version,
     revision: String,
     r#override: Option<PathBuf>,
+}
+
+impl LockSourceRepository {
+    pub fn url(&self) -> &UrlPath {
+        &self.url
+    }
+
+    pub fn path(&self) -> &Path {
+        &self.path
+    }
+
+    pub fn project(&self) -> &str {
+        &self.project
+    }
+
+    pub fn version(&self) -> &Version {
+        &self.version
+    }
+
+    pub fn revision(&self) -> &str {
+        &self.revision
+    }
+
+    pub fn local_path(&self) -> Result<PathBuf, MetadataError> {
+        Ok(Lockfile::dependency_path(&self.url, &self.path, &self.revision)?.join(&self.path))
+    }
 }
 
 impl PartialOrd for LockSource {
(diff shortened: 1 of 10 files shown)


crates/metadata/src/lockfile.rs
```diff
@@ -87,6 +101,32 @@ pub struct LockSourceRepository {
     r#override: Option<PathBuf>,
 }
 
+impl LockSourceRepository {
+    pub fn url(&self) -> &UrlPath {
```
dalance: I think these fields can be `pub` instead of getter method.

dalance (commented):
There are hardly any issues, but I have left one comment. Also, rebasing onto the latest `master` would allow for a smooth merge.

crates/metadata/src/lockfile.rs
pxlsy2000: Thanks, that makes sense. I updated `url`, `path`, `project`, `version`, and `revision` to public fields and removed the corresponding getter methods.

dalance (approved):

PR #2935, opened 2026-06-30
Support Cargo-like external subcommands

### LINKED ISSUE #2933: Support Cargo-like external subcommands
## Motivation

Issue #2648 discusses exposing stable metadata so external integration tools can reuse Veryl dependency resolution. A complementary integration point would be a Cargo-like external subcommand mechanism, so external tools can provide commands that feel native to the Veryl workflow without moving project-specific EDA/IP logic into Veryl core.

For example, a separately installed executable could provide commands such as:

```console
veryl import
veryl flist
```

while Veryl itself only owns the generic dispatch behavior.

## Proposed behavior

- If a requested subcommand is not built into Veryl, search `PATH` for an executable named `veryl-<subcommand>`.
- Dispatch `veryl <subcommand> ...` to that executable while preserving the remaining arguments.
- Preserve the external command exit status.
- Report an actionable diagnostic when the external executable is missing.
- Keep external command implementations outside the Veryl source tree.

It would also be useful for root help to list discoverable external subcommands from `PATH`, optionally with a short description supplied by adjacent metadata from the external tool.

## Scope

This issue is only about the generic Veryl CLI integration surface. It should not require Veryl to understand any external tool-specific metadata, filelist generation policy, import behavior, or downstream EDA tool syntax.

Related: #2648

pxlsy2000:
## Summary
- dispatch unknown Veryl subcommands to `veryl-<subcommand>` executables found on `PATH`
- preserve forwarded arguments and external command exit status
- list discoverable external subcommands in root help, including optional sidecar descriptions

## Testing
- `cargo fmt --all -- --check`
- `cargo test -p veryl --test external_subcommand --test help_external_subcommand`

Closes #2933

### DIFF
--- a/crates/veryl/src/main.rs
+++ b/crates/veryl/src/main.rs
@@ -6,8 +6,6 @@ use fern::Dispatch;
 use log::debug;
 use log::{Level, LevelFilter};
 use miette::{IntoDiagnostic, Result};
-use std::ffi::{OsStr, OsString};
-use std::io::Write;
 use std::process::ExitCode;
 use veryl_metadata::Metadata;
 
@@ -21,11 +19,6 @@ static GLOBAL: mimalloc::MiMalloc = mimalloc::MiMalloc;
 // ---------------------------------------------------------------------------------------------------------------------
 
 fn main() -> Result<ExitCode> {
-    if root_help_requested(std::env::args_os()) {
-        print_root_help()?;
-        return Ok(ExitCode::SUCCESS);
-    }
-
     let opt = Opt::parse();
 
     if opt.list {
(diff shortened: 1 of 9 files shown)


dalance:
I have a few comments regarding the design.

First, since discovering subcommands may be relatively costly (especially in environments with many paths),
how about full list is shown only when `--list` option is used like cargo,
and have `help`, `-h`, and `--help` display something like this:

```
    install     Install a Rust binary
    uninstall   Uninstall a Rust binary
    ...         See all commands with --list
```

Also, requiring the placement of `*.help.toml` files is not a very standard installation practice, so it might confuse users.
How about accepting the performance cost when running `--list` and instead retrieving descriptions by executing the binary itself (e.g., `veryl-import --info`)?

Since these are UX-related issues, I am also fine with merging this as-is for now and considering these points later.

dalance:
Additionally, `veryl-ls` will be detected as external subcommand, but it can't be used as subcommand.
So could you exclude it explicitly?

pxlsy2000:
Thanks for the feedback. I agree with your proposed direction.

Since this affects the CLI interface, I think it is better to settle the behavior before merging. I'll update the help mechanism so full external subcommand discovery is done only for an explicit `--list` flow, and revise the description mechanism along the lines you suggested instead of using `*.help.toml`.

 I'll also explicitly exclude `veryl-ls`, since it is a language server binary rather than a `veryl <subcommand>` style command. For now, I'll keep the external command naming convention Cargo-like (`veryl-<name>`), and handle official non-subcommand binaries through an explicit reserved-name list.

crates/veryl/src/main.rs
```diff
@@ -137,3 +165,23 @@ fn main() -> Result<ExitCode> {
         Ok(ExitCode::FAILURE)
     }
 }
+
+fn root_help_requested(args: impl IntoIterator<Item = OsString>) -> bool {
```
dalance: How about using `#[command(after_help = "\n... See all commands with --list")]` instead of the hand written help override?

pxlsy2000: Hi @dalance 
Agree, I've update the branch, please have a check, thanks

dalance:
Thanks!
I'll merge this PR.

PR #2951, opened 2026-07-01
fix: unkind mismatch_type message for modport port

### LINKED ISSUE #2938: Unkind error message for `mismatch_type` error
Error message for `mismatch_type` error reported from port delcartion item is unkind.

Example:

```systemverilog
interface a_if {
    var a: logic;
    modport mp {
        a: input,
    }
}

module b_module (
    a: modport a_if,
) {
}
```

output
```console
]8;;https://doc.veryl-lang.org/book/07_appendix/02_semantic_error.html#mismatch_type\mismatch_type (link)]8;;\

  💥 "a" is expected to "modport", but it is "modport"
    ╭─[:9:5]
  8 │ module b_module (
  9 │     a: modport a_if,
    ·     ┬
    ·     ╰── Error location
 10 │ ) {
    ╰────
  help: 
```

For the above case, `a_if` is an interface but not `modport` so the error message should be refine.

ranaumarnadeem:
fix  veryl-lang/veryl#2938

When a module port is declared `modport <name>` but `<name>` resolves to an interface rather than a modport of that interface, the reported `mismatch_type` error is confusing:

```
"a" is expected to "modport", but it is "modport"
```

Both sides read the same, so the message gives no hint about what's actually wrong.

This happens because the error's `actual` field was built from the port symbol itself (`Port { direction: Modport }`, which always stringifies to `"modport"` via `to_kind_name()`), instead of from the resolved type of the declared name. The resolved type (`r#type.kind`) already correctly identifies `a_if` as an interface a few lines earlier — it's just never consulted for the error. `TypeKind` already has a `Display` impl used elsewhere in the analyzer for the same kind of message (e.g. `"instance a_if"`, `"modport a_if::<32>::mp"`), so reusing it instead of `to_kind_name()` is enough to fix this:

```
"a" is expected to "modport", but it is "interface a_if"
```

I've added test coverage for this, though since it's a narrow, one-line message-formatting fix in an otherwise well-covered code path, the tests aren't strictly load-bearing — happy to drop them from this PR if you'd rather keep the diff minimal.


dalance:
I'll merge this PR.

PR #2968, opened 2026-07-03
Introduce `project property` feature

### LINKED ISSUE #554: [Feature] Global properties like `define macro`
SV's `define macro` is a useful feature for following porposes.

* Sharing properties globally
* Setting properties from external

I'd like Veryl to have such feature natively instead of preprocessor.

[Pkl](https://github.com/apple/pkl) has an accessor method to get env variables, properties set via the CLI switch, etc.
I prefer such feature.
https://pkl-lang.org/main/current/language-reference/index.html#resources

taichi-ishitani:
close veryl-lang/veryl#554

### DIFF
--- a/Veryl.lock
+++ b/Veryl.lock
@@ -14,6 +14,8 @@ project = "veryl_sample"
 version = "0.16.0"
 revision = "df08333919cdec117aaa699ec97c5c9a2dee69b8"
 
+[projects.properties]
+
 [[projects]]
 name = "veryl_sample2"
 dependencies = []
@@ -25,6 +27,8 @@ path = ""
 project = "veryl_sample"
 version = "0.17.0"
 revision = "2f547db8a5d7c6273bf84ebbe540e315e84fe47f"
+
+[projects.properties]
 
 [[projects]]
 name = "veryl_sample3"
@@ -38,7 +42,13 @@ project = "veryl_sample"
 version = "0.18.0"
 revision = "52fc06237c5b70d42d67ea0910874ba111a6ef98"
 
+[projects.properties]
+
 [[projects]]
 name = "veryl_sample4"
 source = "testcases/sample"
 dependencies = []
+
+[projects.properties]
+QUX_0_VALUE = 16
+QUX_1_VALUE = false
--- a/crates/analyzer/src/conv/utils.rs
+++ b/crates/analyzer/src/conv/utils.rs
@@ -2564,6 +2564,20 @@ pub fn eval_factor_symbol(
 
             return Ok(ir::Factor::Value(x));
         }
+        SymbolKind::ProjectProperty(x) => {
+            let value = match &x.value {
+                veryl_metadata::ProjectProperty::Int(x) => Value::new(*x as u64, 64, true),
+                veryl_metadata::ProjectProperty::Bool(x) => {
+                    if *x {
+                        Value::new(1, 1, false)
+                    } else {
+                        Value::new(0, 1, false)
+                    }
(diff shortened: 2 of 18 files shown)


crates/analyzer/src/conv/utils.rs
```diff
@@ -2564,6 +2564,20 @@ pub fn eval_factor_symbol(
 
             return Ok(ir::Factor::Value(x));
         }
+        SymbolKind::ProjectProperty(x) => {
+            let value = match &x.value {
+                veryl_metadata::ProjectProperty::Int(x) => Value::new(*x as u64, 64, true),
+                veryl_metadata::ProjectProperty::Bool(x) => {
+                    if *x {
+                        Value::new(1, 64, false)
```
dalance: Is it better to use `1` as bit width instead of `64`?

crates/metadata/src/lockfile.rs
dalance: `add` becomes to checking by `x.uuid == lock.uuid`, but remove is still `x.source == old_lock.source`.
Should this be based on uuid too?

crates/metadata/src/lockfile.rs
```diff
@@ -37,8 +37,12 @@ pub struct Lockfile {
 pub struct Lock {
     pub name: String,
     pub source: LockSource,
+    #[serde(default)]
+    pub properties: HashMap<String, ProjectProperty>,
```
dalance: The order of `HashMap` is random.
So serialization result becomes random.
How about using `BTreeMap` instead of it?

crates/metadata/src/lockfile.rs
```diff
@@ -37,8 +37,12 @@ pub struct Lockfile {
...
     pub dependencies: Vec<LockDependency>,
     #[serde(skip)]
+    pub uuid: Uuid,
```
dalance: `uuid` is skipped by serde, so if there is a path to forget uuid assignment, the uuid becomes invalid.
Instead of it, how about introducing `uuid()` method to calculate it from other fields?

crates/analyzer/src/conv/utils.rs
taichi-ishitani: This is simply my mistake. The width should be 1.

crates/metadata/src/lockfile.rs
taichi-ishitani: Yes, it should be also based on UUID.

crates/metadata/src/lockfile.rs
taichi-ishitani: I introduced `fn uuid` instead of the `uuid` field.

crates/metadata/src/lockfile.rs
taichi-ishitani: I changed types of `Metadata::properties` and `Lock::properties` to `BTreeMap`.

taichi-ishitani:
@dalance 
I applied your feedbacks. Can you review the PR again?

Veryl.lock
```diff
@@ -38,7 +42,13 @@ project = "veryl_sample"
...
+[projects.properties]
+
 [[projects]]
 name = "veryl_sample4"
 source = "testcases/sample"
 dependencies = []
+
+[projects.properties]
+QUX_1_VALUE = false
+QUX_0_VALUE = 16
```
dalance, line 53: This order may be incorrect.
Could you recreate Veryl.lock?

taichi-ishitani: I created it again.

PR #2969, opened 2026-07-03
Fix interface parameter overrides dropped through modport ports

shuntarot:
A module receiving an interface through a modport port was elaborated once and cached keyed only on its own parameters, so the first elaboration (e.g. against a default-parameter interface) was reused for instances connected to interfaces with overridden parameters:

    interface If #( param DW: u32 = 64 ) { var data: logic<DW>; ... }
    inst a_if: If;                 // DW=64
    inst b_if: If #( DW: 128 );
    inst a: Consumer ( xif: a_if );
    inst b: Consumer ( xif: b_if ); // reused DW=64 body: data truncated

Fold the connected interface instances' signatures into the consuming module's Signature (participating in the instance-cache key and normalize()), and register modport ports' effective interface signatures so pass-through connections (module A's modport port forwarded to a child's modport port) key correctly as well. Adds the interface_parameter_override simulator regression test (direct and pass-through modports, default + overridden widths).

### DIFF
--- a/crates/analyzer/src/conv/context.rs
+++ b/crates/analyzer/src/conv/context.rs
@@ -478,7 +478,10 @@ impl Context {
         self.inst_signatures.insert(identifier.text(), sig.clone());
     }
 
-    pub fn collect_modport_signatures(&mut self, inst: &ComponentInstantiation) {
+    pub fn collect_modport_signatures(
+        &mut self,
+        inst: &ComponentInstantiation,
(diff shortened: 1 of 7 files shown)


dalance:
Could you fix the clippy error?

shuntarot:
Yes, let me check

dalance:
I'll merge this PR.

Issue #2891, opened 2026-06-24
Add support for recursive generic functions


coolshaurya:
Add support for recursive generic functions. As of Veryl 0.20.1, recursive generic functions get compiled by the Veryl compiler to invalid SystemVerilog where the nested calls to the same function don't get generated.

That is, if the recursive function `function y<WIDTH: p32> (...)` calls `y::<WIDTH - 1>(...)` and the top-level call is `y::<4>(...)` then only `__y__4` that calls `__y__3` is generated but `__y__3` is not generated, causing an error when simulating the generated SV.

<details>
<summary>
Practical code example of this feature being used to count leading zeroes
</summary>

```
// Method taken from https://ieeexplore.ieee.org/document/10738924

package count_leading_zeroes_ {

    function clz_8 (
        a: input logic<8>,
    ) -> logic<3> {
        var out     : logic<3>;
        var z2a     : logic<2>;
        var z2b     : logic<2>;
        let a_inv   : logic<8> = ~a;
        out[2]   = &a_inv[7:4];
        z2a      = {&a_inv[7:6], |a_inv[6:5] & a_inv[7]};
        z2b      = {&a_inv[3:2], |a_inv[2:1] & a_inv[3]};
        out[1:0] = if out[2] ? z2b : z2a;
        return out;
    }

    function clz::<OUT_WIDTH: u32> (
        a: input logic<(1 << (OUT_WIDTH - 1))>,
    ) -> logic<OUT_WIDTH> {
        let out: logic<OUT_WIDTH> = clz_rec::<OUT_WIDTH>(a);
        return if out[msb] ? (1 << (OUT_WIDTH - 1)) : out;
    }

    function clz_rec::<OUT_WIDTH: u32> (
        a: input logic<(1 << (OUT_WIDTH - 1))>,
    ) -> logic<OUT_WIDTH> {
        gen IN_WIDTH     : u32 = (1 << (OUT_WIDTH - 1));
        gen HALF_IN_WIDTH: u32 = IN_WIDTH >> 1;
        gen OUT_WIDTH_DEC: u32 = OUT_WIDTH - 1;

        var out  : logic<OUT_WIDTH>;
        let a_inv: logic<IN_WIDTH>  = ~a;

        out[msb] = &a_inv;
        if OUT_WIDTH <: 4 {
            $fatal();
        } else if OUT_WIDTH == 4 {
            out[2:0] = clz_8(a);
        } else {
[...]

PR #2971, opened 2026-07-03
fix(analyzer): report infinite_recursion instead of crashing on recur…

ranaumarnadeem:
Recursive generic functions crash the compiler with a stack overflow (#2891).
Evaluating a generic function call expands the callee's body at compile time,
which contains another call to itself (`f::<4>` → `f::<3>` → …), recursing until
the stack is exhausted and the process aborts.

This adds a depth guard in `eval_factor_path` so we report `infinite_recursion`
and stop instead of crashing. Because that evaluation runs on an error-tolerant
path, the hit is recorded and surfaced from `create_ir`. Modules already get
similar protection via `push_instance_history`; functions had none.

It's a stopgap so the compiler errors cleanly and does not crash
It isnot the recursive-generics feature itself, about which I am still thinking how to do that.

### DIFF
--- a/crates/analyzer/src/analyzer.rs
+++ b/crates/analyzer/src/analyzer.rs
@@ -128,14 +128,22 @@ impl Analyzer {
     fn create_ir(context: &mut Context, input: &Veryl) -> (Ir, Vec<AnalyzerError>) {
         let ir: IrResult<Ir> = Conv::conv(context, input);
 
-        if let Ok(mut ir) = ir {
+        let (ir, mut errors) = if let Ok(mut ir) = ir {
             ir.eval_assign(context);
             let errors = context.drain_errors();
             (ir, errors)
         } else {
             let errors = context.drain_errors();
             (Ir::default(), errors)
+        };
+
+        // Surface a generic-evaluation recursion-limit hit (#2891) that the
+        // error-tolerant evaluation path would otherwise swallow.
+        if let Some(token) = crate::conv::utils::take_generic_eval_overflow() {
+            errors.push(AnalyzerError::infinite_recursion(&token));
         }
+
+        (ir, errors)
     }
 
     pub fn analyze_pass2(
--- a/crates/analyzer/src/analyzer.rs
+++ b/crates/analyzer/src/analyzer.rs
@@ -165,10 +165,7 @@ impl Analyzer {
             (Ir::default(), errors)
         };
 
-        // Surface a function-eval recursion-limit hit (#2891) that the
-        // error-tolerant evaluation path would otherwise swallow. It is reported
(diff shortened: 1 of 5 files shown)


dalance:
For instance, infinite recursion detection for instances is handled by maintaining an `InstanceHistory` within the `Context` and using `push_instance_history`.
I believe it is cleaner to store this information in that structure rather than using TLS.
Also, simply using a threshold of 16 would result in finite but deep structures being incorrectly identified as infinite recursion.
Since `push_instance_history` distinguishes between infinite recursion and exceeding a finite depth limit,
it would be ideal to implement the detection in a similar way.

ranaumarnadeem:
@dalance  Agreed on both counts moving the recursion state into Context.instance_history and reusing push_instance_history is cleaner than the thread-local.. I'll rework it to reuse InstanceHistory so a genuinely cyclic case eports InfiniteRecursion, while a merely-too-deep case reports ExceedDepthLimit 

One thing I ran into that I'd like your call on the function comptime-evaluation path (eval_factor_path → eval_factor_symbol → eval_generic_expr → …) is much more stack-hungry per level than module instantiation — ~14 frames/level, and veryl check overflows the native stack at ~depth 33 on the default 8 MB stack (that's why the stopgap's threshold was so low). So with the module instance_depth_limit of 1024, the native stack still overflows before ExceedDepthLimit fires. Reaching 1024 would need a ~250 MB stack.

My proposal is wire push_instance_history into the function path for the correct InfiniteRecursion-vs-ExceedDepthLimit distinction, but give function-eval recursion its own stack-safe depth limit (e.g. ~128–256, still far above realistic recursive generics) reported as ExceedDepthLimit. Longer term the right fix is making that evaluation less deeply recursive so it can share the 1024 limit. Are you OK with a separate, lower function-eval depth limit? or we should just optimizine to reduce stack depth ?

dalance:
I agree "a separate, lower function-eval depth limit".

dalance:
Thank you.
I think the detection logic is fine.

There are quite a lot of comments overall, so it would be best to adjust that aspect.
For instance, the meaning of `function_instance_depth_limit` is obvious from its name,
and since other entries don't have doc comments, this one doesn't need them either.

Excessive comments make the source code harder to read,
and background information, such as references to issue numbers, can be traced via `git blame` and PRs,
so there is no need to include them as comments.

I believe most of the comments are unnecessary or could be condensed to just one or two lines.

ranaumarnadeem:
@dalance apologies i pushed the wrong one less commented version is still in my local i will fix and push it

dalance:
Looks good!
I'll merge this PR.

ranaumarnadeem:
Thanks I will try to implement the proper feature this weekend

PR #2975, opened 2026-07-05
fix(analyzer): emit mismatch_type for non-FactorType proto generic default

### LINKED ISSUE #2382: No error is reported for invalid default value assignment to `proto` generic param
`mismatch_type` error should be reported for the following code but not.

```systemverilog
proto module ProtoModuleA;
module ModuleA::<M: ProtoModuleA = 1> {}
```

mvanhorn:
## Summary

Proto generic parameters now reject a number/boolean default value unless the bound resolves to a variable (factor) type. A default like `M: ProtoModuleA = 1` is reported as a `mismatch_type` error, while a variable-type bound such as `M: u32 = 1` still analyzes clean.

## Why this matters

Reported in #2382. The `Conv<&WithGenericParameterItem>` impl in `crates/analyzer/src/conv/declaration.rs` guarded the numeric/boolean default arm with `if !matches!(x.bound, GenericBoundKind::Proto(_))`, which blanket-accepted a numeric default for every proto bound. For a `proto module` (or `proto interface` / `proto package`, and enum/struct/union type bounds) a number is not a valid default, so no error was emitted for the reproduction:

```veryl
proto module ProtoModuleA;
module ModuleA::<M: ProtoModuleA = 1> {}
```

The fix resolves the bound with `resolve_proto_bound` and only accepts the number when it is a `ProtoBound::FactorType`, otherwise emitting `mismatch_type` with `MismatchTypeKind::GenericArgument`. This mirrors the existing call-site check in `check_generic_proto_arg` (`crates/analyzer/src/conv/checker/generic.rs`), which already rejects a non-component argument for those bounds.

## Testing

- Added subtests to the existing `mismatch_type` test in `crates/analyzer/src/tests.rs`: the issue's `proto module` reproduction now yields a `MismatchType` error, and the `M: u32 = 1` regression happy path still analyzes clean.
- `cargo test -p veryl-analyzer mismatch_type` passes.
- `cargo fmt --check` and `cargo build` are clean.

Fixes #2382

### DIFF
--- a/crates/analyzer/src/conv/declaration.rs
+++ b/crates/analyzer/src/conv/declaration.rs
@@ -26,7 +26,7 @@ use crate::ir::{
 };
 use crate::namespace::DefineContext;
 use crate::symbol::{
-    ClockDomain, Direction, GenericBoundKind, Symbol, SymbolKind, TbComponentKind,
+    ClockDomain, Direction, GenericBoundKind, ProtoBound, Symbol, SymbolKind, TbComponentKind,
 };
 use crate::symbol_path::{GenericSymbolPath, SymbolPathNamespace};
 use crate::symbol_table;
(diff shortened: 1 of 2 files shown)


mvanhorn:
Thanks @dalance! Glad proto generic args now surface mismatch_type instead of slipping through.

mvanhorn:
Appreciate it as always, @dalance! Good to have mismatch_type emitted for non-FactorType proto generic defaults.

PR #2996, opened 2026-07-13
Introduce `mixin` declaration to interface

### LINKED ISSUE #1212: Feature to share implementation
For [pzbcm](https://github.com/pezy-computing/pzbcm), I use text macros to share implementation.
For example:

* https://github.com/pezy-computing/pzbcm/blob/985e7147ab42682e62f32604433090b3be5f8482/pzcorebus_common/pzcorebus_if_internal_macros.svh#L10
* https://github.com/pezy-computing/pzbcm/blob/985e7147ab42682e62f32604433090b3be5f8482/pzcorebus_common/pzcorebus_request_if.sv#L18
* https://github.com/pezy-computing/pzbcm/blob/985e7147ab42682e62f32604433090b3be5f8482/pzcorebus_common/pzcorebus_if.sv#L18

I need a new feature to share implementation between design elements to re-write these interfaces in Veryl.
Followings are idenas for this.

* Unnamed code block.

This is similar to #793 but no instance name is needed.
Code block will be expanded into the scope where it is instantiated.

* Interface inheritance

For the above example, `pzcorebus_if` behaves `pzcorebus_request_if` and also `pzcorebus_response_if`.
Therefore, I think `inheritance` is one of suitable ways to model the above case.
Implementations of parent interfaces are expanded into the child interface.

taichi-ishitani:
Close #1212

Adds interface mixin via a new `mixin` declaration. Inside an interface, `mixin <interface>;` mixes in the members (variables, functions, and modports) of another interface; the mixed-in members are expanded into the interface as if they were declared directly.

## Syntax

`mixin` is an interface declaration item:

```veryl
interface InterfaceC {
    mixin InterfaceA;
    mixin InterfaceB;

    // additional members of the interface
}
```

Generic arguments can be passed to a mixed-in interface:

```veryl
interface InterfaceC::<W: u32> {
    mixin InterfaceA::<W>;

    // ...
}
```

## Restrictions on a mixed-in interface

A mixed-in interface must satisfy all of the following; otherwise an `InvalidMixin` error is reported:

* It must be an interface, not a `proto` interface.
* It must not have parameters (generic parameters are allowed).
* It must not itself mix in other interfaces (nested mixin is not supported yet).
* Its member names must not conflict with the interface's own members or with the other mixed-in interfaces' members.

## Generated SystemVerilog

The mixed-in members are expanded inline into the interface. Given:

```veryl
package DataPkg::<W: u32> {
    type Data = logic<W>;
}

interface Req::<W: u32> {
    import DataPkg::<W>::*;
    var command: Data;
    modport mp_req {
        command: output,
    }
}

interface Rsp {
    var status: logic;
    modport mp_rsp {
        status: input,
    }
}

interface Bus::<W: u32> {
    mixin Req::<W>;
    mixin Rsp;

    modport mp {
        ..same(mp_req, mp_rsp)
    }
}
```

`Bus` is emitted with the mixed-in members expanded (names simplified for readability):

```systemverilog
interface Bus__W;
    // mixed in from Req
    DataPkg__W::Data command;
    modport mp_req (output command);

    // mixed in from Rsp
    logic status;
    modport mp_rsp (input status);

    // the interface's own member
    modport mp (
        output command,
        input  status
    );
endinterface
```

Notes:

* A member referenced through a mixed-in interface's `import` is emitted in package-qualified form (e.g. `DataPkg__W::Data`) instead of merging the `import` statements. This avoids conflicts when mixed-in interfaces import same-named symbols from different packages.
* Generic arguments passed to a mixed-in interface are propagated into the expanded members.


dalance:
The whole changes look good.
Could you add a snapshot test for `InvalidMixin` error?

crates/analyzer/src/symbol_table.rs
```diff
@@ -23,6 +24,7 @@ use log::trace;
 use msb::check_msb;
 use serde::{Deserialize, Serialize};
 use std::cell::RefCell;
+use std::collections::HashSet;
```
dalance: Could you use `crate::HashSet` instead of std?

taichi-ishitani: I changed it to `crate::HashSet`.

taichi-ishitani:
> The whole changes look good. Could you add a snapshot test for `InvalidMixin` error?

I added snapshot tests for `InvalidMixin` error.

PR #3078, opened 2026-07-22
Add `$tb::random` testbench component

taichi-ishitani:
Adds `$tb::random`, which lets test scenarios generate random numbers in native tests (`veryl test`). Like `$tb::file`, it is an object-style component; the type of the generated value is given as a generic argument.

## Usage

Declare it as a `var` inside a `#[test]` module, giving the element type as a generic argument.

```systemverilog
#[test(test_rand)]
module test_rand {
    var r: $tb::random::<u32>;
    var x: u32;

    initial {
        r.seed(42);                 // set the seed
        x = r.get();                // uniform random over the full u32 range
        x = r.get_range(10, 20);    // 10 <= x <= 20 (like $urandom_range)
        let s: u64 = r.get_seed();  // read back the current seed
        $finish();
    }
}
```

### Methods

| Method | Description |
|---|---|
| `seed(value)` | Set the seed. |
| `get()` | Generate a uniform random value over the full range of the element type `T` (return type = `T`). |
| `get_range(min, max)` | Generate a uniform random value in `min <= x <= max` (inclusive; like `$urandom_range`). |
| `get_seed()` | Read back the current seed (64-bit). |

`get()` / `get_range()` / `get_seed()` return a value, so they can be used in assignment, expression, and typed-`let` positions.

```systemverilog
let y: u32 = r.get() + 1;   // the return type T is resolved inside expressions too
```

### Allowed types

- `u8`–`u64` / `i8`–`i64` (signed types produce signed values), and `bbool`.
- Width must be **at most 64 bits**. `bit<N>` (N ≤ 64) is also usable via a `gen` type binding or a package type:

```systemverilog
gen my_t: type = bit<12>;
var r: $tb::random::<my_t>;   // 12-bit random
```

- Non-integer (`f32`/`f64`), 4-state (`lbool`/`logic`), and widths over 64 bits are compile errors.

## Reproducibility

Giving a seed via `--seed` or `[test].seed` in `Veryl.toml` makes the generated sequence reproducible **within the same Veryl binary**.

- Without an explicit seed, each run is random (the seed used is logged, so a run can be reproduced with `--seed`).
- Calling `r.seed(v)` pins a specific object's sequence.
- Even without an explicit seed, each object is initialized with a seed derived deterministically from the test name and the object name, so the same `--seed` yields the same sequence every run (the same scheme already used for component randomness).

> Note: long-term reproducibility across `rand` versions is out of scope (only `--seed` reproducibility within the same binary is guaranteed).

## Implementation notes

- The PRNG uses the `rand` crate (`StdRng` + `random_range`). It is added with `default-features = false` so the wasm build (playground) is not broken.
- Like `$tb::file`, it can only be used inside a `#[test]` module and is never emitted to SystemVerilog.
- The element type `T` is modeled as a generic parameter of `$tb::random`, so `get()`'s return type `T` is resolved statically through the existing generic-resolution machinery.
- As a side fix, this also fixes an `eval_type` gap where the width was lost when a `gen` type binding (`gen w: type = bit<N>`) was passed as a generic argument (`VariableType` was not handled).

## Tests

- analyzer: valid usage (multiple element types, all methods, return-type resolution); rejection of invalid element types.
- simulator: runtime determinism and range; generation for `bit<N>` reached via `gen`.

### DIFF
--- a/crates/simulator/src/random_table.rs
+++ b/crates/simulator/src/random_table.rs
@@ -1,0 +1,137 @@
+//! Thread-local RNG table backing the `$tb::random` testbench handles.
+//!
+//! Like `file_table`, a thread-local keeps the generator state reachable from
+//! the testbench driver (which has no `Simulator` handle) and isolates
+//! parallel tests. Generators are keyed by the declaring variable's name
+//! (`StrId`). Each handle is lazily seeded from the run's base seed (`--seed`
+//! / `[test].seed`, i.e. `Ir.seed`) mixed with the handle name, so a run is
+//! reproducible for a given seed and distinct handles get distinct streams.
+//! `seed()` overrides the seed explicitly; `get_seed()` reads it back.
+
+use crate::ir::Value;
+use rand::{RngExt, SeedableRng};
+use rand_pcg::Pcg64;
+use std::cell::RefCell;
+use std::collections::HashMap;
+use veryl_parser::resource_table::{self, StrId};
+
+#[derive(Default)]
+struct RandomTable {
+    /// Run base seed (`Ir.seed`); handles derive their seed from it.
+    base_seed: u64,
+    /// Per-handle generator paired with the seed currently applied to it.
+    /// A non-cryptographic PCG generator is used (fast and reproducible; the
+    /// testbench does not need a crypto-strength RNG).
+    rngs: HashMap<StrId, (Pcg64, u64)>,
+}
+
+thread_local! {
+    static TABLE: RefCell<RandomTable> = RefCell::new(RandomTable::default());
+}
+
+/// Deterministic FNV-1a over the base seed and the handle name; matches the
+/// scheme used for per-instance component seeds.
+fn derive_seed(base: u64, key: StrId) -> u64 {
+    let name = resource_table::get_str_value(key).unwrap_or_default();
+    let mut h: u64 = 0xcbf2_9ce4_8422_2325;
+    let mut eat = |bytes: &[u8]| {
+        for b in bytes {
+            h ^= *b as u64;
+            h = h.wrapping_mul(0x0000_0100_0000_01b3);
+        }
+    };
+    eat(&base.to_le_bytes());
+    eat(name.as_bytes());
+    h
+}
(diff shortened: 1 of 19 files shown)


crates/simulator/src/testbench.rs
```diff
@@ -63,6 +63,29 @@ pub enum TestbenchStatement {
         /// Assignment form: destination variable for the return value.
         ret: Option<(VarId, crate::ir::RetWidthCheck)>,
     },
+    /// `r.seed(v)` — `handle` is the `$tb::random` variable's name (RNG key).
+    RandomSeed { handle: StrId, value: Expression },
+    /// `x = r.generate()`.
```
dalance: `generate` may be old function name?

crates/simulator/src/random_table.rs
```diff
@@ -0,0 +1,135 @@
...
+//! Like `file_table`, a thread-local keeps the generator state reachable from
+//! the testbench driver (which has no `Simulator` handle) and isolates
+//! parallel tests. Generators are keyed by the declaring variable's name
+//! (`StrId`). Each handle is lazily seeded from the run's base seed (`--seed`
+//! / `[test].seed`, i.e. `Ir.seed`) mixed with the handle name, so a run is
+//! reproducible for a given seed and distinct handles get distinct streams.
+//! `seed()` overrides the seed explicitly; `get_seed()` reads it back.
+
+use crate::ir::Value;
+use rand::rngs::StdRng;
```
dalance: How about using `rand_pcg::Pcg64` instead of `StdRng` becuase `StdRng` is crypt-aware and slower than non-crypt rand?

crates/simulator/src/testbench.rs
taichi-ishitani: I will fix it.

crates/simulator/src/random_table.rs
taichi-ishitani: I will fix it.

taichi-ishitani:
I fixed your feedbacks.

PR #3083, opened 2026-07-23
Apply select to a value resolved through the symbol route

### LINKED ISSUE #3082: False positive of `invalid_operand` error
Analyzer reports unexpected `invalid_operand` error for the code below.

```systemverilog
package pkg {
    const WIDTH: u32 = 8;
    const SIZE : u32 = 4;
    const ARRAY: bit<WIDTH> [SIZE] = '{8'h11, 8'h22, 8'h33, 8'h44};

    function f() -> logic<WIDTH> {
        var result: logic<WIDTH>;
        result = '0;
        for i in 0..SIZE {
            result = result + ARRAY[i];
        }
        return result;
    }
}

module ModuleA (
    o_d: output logic<8>,
) {
    assign o_d = pkg::f();
}
```

taichi-ishitani:
Fixes #3082.

## Problem

The analyzer reported a false-positive `invalid_operand` error when a package `const` array is indexed at a reference site:

```veryl
package PkgA {
    const ARRAY: bit<8> [4] = '{8'h11, 8'h22, 8'h33, 8'h44};
}
module ModuleA {
    let _a: bit<8> = PkgA::ARRAY[0];  // ARRAY[0] was typed as bit<8>[4]
}
```

## Cause

`ARRAY` is not registered in the referring scope's variable table, so `find_path` misses and it is resolved through the symbol route (`eval_factor_symbol` / `eval_factor_symbol_external`).

That route takes no `select` argument and silently drops the array index, returning the whole `bit<8>[4]` type instead of the element type `bit<8>`. The array-typed operand then trips the `+` operand check.

### analyzer

`apply_symbol_select` applies the dropped select to whatever `eval_factor_symbol[_external]` returned, branching purely on the select:

| select | result |
| --- | --- |
| none | the `Factor::Value` as-is |
| constant, full index | folds to the selected element (`Factor::Value`) |
| constant, partial index | the selected sub-array, carried as `ValueVariant::NumericArray` |
| dynamic | the new `ir::Factor::SelectedValue` |

Supporting changes:

- `eval_expr` carries an array literal's per-element values as `ValueVariant::NumericArray`, so a later reference can resolve individual elements.
- New `ir::Factor::SelectedValue { values, array, index, select, comptime }` — a const value selected at runtime. It holds the whole-const per-element values inline; `comptime.r#type` is the **post-select** (reduced) type, and `comptime.value` is `Unknown` because the read is dynamic. Constant selection never reaches it (the analyzer folds that above), so the simulator/synthesizer only see it for genuinely dynamic selection.
- The `find_path` branch now clears a whole-array `NumericArray` when the reference applies an index/select — that value describes the whole array, not the selected part, and leaving it there broke sliced parameter overrides (`#(YS: M[1])`).

### simulator

New `ProtoExpression::DynamicValue` / `Expression::DynamicValue`, the value-side counterpart of `DynamicVariable`: it reads an element by runtime index from an **inline constant table** instead of a state buffer, then applies the static or dynamic bit-select. Both backends (Cranelift, AOT-C) decline it, so it evaluates on the interpreter.

This keeps the analyzer's `comptime.r#type` at the correct post-select width instead of widening it to the element width to make a shift/mask lowering work.

### synthesizer

`synth_factor` gains a `SelectedValue` arm: the constant table becomes nets, `arith::dynamic_mux_tree` selects the element, then the bit-select is applied (constant → net slice, dynamic → 1-bit mux). Unsupported cases (multi-dimensional dynamic index, dynamic range select) match the existing `Factor::Variable` path.

## Tests

- `analyzer`: the reported case added to `tests::invalid_operand`; `ir::tests::package_const_select` covers a scalar part-select, an array element, a sub-array (`C[0]`), and a fully-indexed element (`C[1][1]`).
- `simulator`: `tests::simulation::package_const_select` checks static part-select / static index / dynamic part-select / dynamic index across all configs (interpreter, JIT, AOT-C, 4-state).
- `synthesizer`: `package_const_dynamic_select` checks that a dynamic const-array read synthesizes to a one-hot decode, that an all-identical table folds to a constant, and that `4'b1010[sel]` reduces to `sel[0]`.

Full workspace test suite passes.

## Notes

**Signedness of `u32` const array elements.** `ir::tests::array_literal` now expects `32'h0000000a` where it used to expect `32'sh0000000a`. Carrying the element values on the comptime moves a const array literal from the `ArrayLiteral` branch of `eval_const_assign` to the `NumericArray` branch, which applies `set_signed(r#type.signed)` via `fit_array_elements`. If the stored values represent "the value held by `const X: u32[3]`", propagating the declared type's signedness looks right, but the literal itself is an `i32` array — **is this propagation intended?**

### DIFF
--- a/crates/analyzer/src/conv/utils.rs
+++ b/crates/analyzer/src/conv/utils.rs
@@ -183,7 +183,7 @@ pub fn eval_expr(
     if let Ok(mut expr) = expr {
         check_anonymous(context, &expr, allow_anonymous, token);
 
-        let comptime = if let Some(dst_type) = dst_type {
+        let mut comptime = if let Some(dst_type) = dst_type {
             let mut comptime = expr.eval_comptime(context, dst_type.total_width()).clone();
 
             check_compatibility(context, &dst_type, &comptime, &token);
@@ -193,6 +193,15 @@ pub fn eval_expr(
         } else {
             expr.eval_comptime(context, None).clone()
         };
+
+        // `eval_comptime` doesn't evaluate an array literal, so evaluate it here
+        // to carry the element values.
+        if matches!(expr, ir::Expression::ArrayLiteral(..))
+            && !comptime.r#type.array.is_empty()
+            && let Some(values) = eval_array_literal_values(context, &comptime.r#type, &mut expr)?
+        {
+            comptime.value = ValueVariant::NumericArray(values);
+        }
 
         Ok((comptime, expr))
     } else {
@@ -2234,9 +2243,15 @@ fn eval_factor_path_inner(
             let index = array_select.to_index();
             comptime.r#type.array.drain(0..index.dimension());
 
-            // A const symbol read through a dynamic index/select is not itself
-            // a compile-time constant (e.g. `A[idx]` for a const array `A`).
             comptime.is_const &= index.is_const() && width_select.is_const();
+
+            // The whole-array value doesn't describe a selected part of it; drop
+            // it so consumers resolve the selection from the variable table.
+            if (index.dimension() > 0 || !width_select.is_empty())
+                && matches!(comptime.value, ValueVariant::NumericArray(_))
+            {
+                comptime.value = ValueVariant::Unknown;
+            }
 
             comptime.token = token;
             if comptime.r#type.is_type() {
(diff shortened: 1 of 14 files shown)


dalance:
Overall, the comments tend to be lengthy, and some simply restate what is already evident from the code itself.
Excessive commenting can make the code harder to read and maintain; I suggest being more concise and focusing on information that cannot be gleaned directly from the code.

Also, regarding the implementation that falls back to the interpreter during simulation:
The interpreter is not only more than 100 times slower than Cranelift/AOT-C, but it also incurs a massive performance penalty because the entire process falls back to it if even a single `DynamicValue` is present.
Is it possible to include the Cranelift/AOT-C implementation as well?

As for `ir::tests::array_literal`, I believe the new expected value is correct.

taichi-ishitani:
> comments 

I removed unnecessary comments.

> Cranelift/AOT-C

I'm not able to judge the correctness of the implementation because I have no background knowledge. I'd like you to implement this.

PR #3107, opened 2026-07-25
Fix `unknown_member` error reported for items imported from generic packages defined in other projects

### LINKED ISSUE #3104: Unknown member error for item imported from generic package defined in other project
An `unknown member` error is reported for an item imported from a generic package defined in an other project.

Example:

a.veryl
```systemverilog
package a_pkg {
  const WIDTH: u32 = 32;
}
```

b.veryl
```systemverilog
module b_module {
  import a_pkg::*;
  import $std::lzc_pkg::<WIDTH>::lzc_result;
  import $std::lzc_pkg::<WIDTH>::lzc;
  let _result: lzc_result = lzc(32'h0123_4567, true);
}
```

```console
Error: unknown_member (link)

  × "$std" doesn't have member "lzc_result"
   ╭─[/home/taichi/workspace/temp/veryl/71/b.veryl:5:16]
 4 │   import $std::lzc_pkg::<WIDTH>::lzc;
 5 │   let _result: lzc_result = lzc(32'h0123_4567, true);
   ·                ─────┬────
   ·                     ╰── Error location
 6 │ }
   ╰────
  help:

Error: unknown_member (link)

  × "$std" doesn't have member "lzc"
   ╭─[/home/taichi/workspace/temp/veryl/71/b.veryl:5:29]
 4 │   import $std::lzc_pkg::<WIDTH>::lzc;
 5 │   let _result: lzc_result = lzc(32'h0123_4567, true);
   ·                             ─┬─
   ·                              ╰── Error location
 6 │ }
   ╰────
  help:
```

taichi-ishitani:
fix veryl-lang/veryl#3104

Prefix paths inserted by `GenericSymbolPath::resolve_imported` have been skipped in
`ReferenceTable::generic_symbol_path` because they were assumed to be resolved already.
However that loop does not only resolve paths but also prepends the project name to generic
arguments. Therefore the generic argument of a skipped prefix path was left without the
project name.

For example, with the testcase added by this PR:

```veryl
alias package bazpkg = veryl_sample4::baz_pkg::<Package25::<1>::C>;

module Module25F {
    import bazpkg::BAZ;
    let _f: u32 = BAZ;
}
```

the path of `BAZ` at the `let` declaration has been resolved as

```
veryl_sample4::baz_pkg::<Package25::<1>::C>::BAZ
```

but it should be resolved as

```
veryl_sample4::baz_pkg::<veryl_testcase::Package25::<1>::C>::BAZ
```

The former does not match the token of the actual generic instance of `baz_pkg`,
so `BAZ` cannot be found and the `unknown_member` error is reported.

To fix this, prefix paths should be resolved too.

----

Resolving prefix paths reveals another problem.
`GenericSymbolPath::resolve_imported` unaliases the inserted package path, but it does not have
the generic maps of the namespace where the alias is declared.
Therefore a generic parameter in the alias target is left unreplaced.

For example, with this code in an existing test:

```veryl
package a_pkg::<a: u32> {
    const A: u32 = a;
}
package b_pkg::<b: u32> {
    alias package a = a_pkg::<b>;
}
module c_module {
    import b_pkg::<32>::a;
    import a::*;
    const C: u32 = A;
}
```

the prefix path inserted for `A` has been unaliased as

```
a_pkg::<b>
```

but it should be

```
a_pkg::<32>
```

`b` is a generic parameter of `b_pkg` and is not visible from `c_module`, so an
`undefined_identifier` error is reported for it.

To fix this, `resolve_imported` should not unalias the package path and should insert it as `a`.
The symbol table resolves an alias transparently, and `ReferenceTable::generic_symbol_path`
expands the alias target with the generic maps collected from the namespace which the alias
belongs to. So `b` is replaced with `32` there.


crates/tests/src/lib.rs
dalance: Should `STD_TESTS` be added to this condition?
Additionally this pattern (`TESTS.contains(name) && TESTS[0] != name`) can be extracted to a helper function.

crates/analyzer/src/reference_table.rs
```diff
@@ -277,14 +277,9 @@ impl ReferenceTable {
...
         let mut generic_maps = generic_maps.cloned().unwrap_or_default();
-
-        let orig_len = path.len();
         path.resolve_imported(scope, define_context, Some(&generic_maps));
 
-        // Prefix paths added by `resolve_imported` have already been resolved.
-        // They should be skipped.
-        let prefix_len = path.len() - orig_len;
-        for i in prefix_len..path.len() {
+        for i in 0..path.len() {
```
dalance, line 285: By this change, `check_package_reference` will be executed for prefix segments too.
So the following code will get 4 referring_before_definition errors ( previously only one).

```
module ModuleA {
    import PackageB::*;
    let _a: u32 = B;
    let _b: u32 = B;
    let _c: u32 = C;
}
package PackageB {
    const B: u32 = 1;
    const C: u32 = 2;
}
```

taichi-ishitani: I understood. Checks for paths added by `impor` should be skipped.

taichi-ishitani:
I fixed these feedbacks.

PR #3118, opened 2026-07-28
fix(ls): add project to the key of per-file table drops

taichi-ishitani:
## Problem

The language server panics during background analysis when the same source directory is referenced by more than one project.

`testcases/25_dependency_2` now reproduces this: `veryl_sample4` and `veryl_sample5` both point at `testcases/sample`, so every file under it is analyzed twice, once per project.

```
thread '<unnamed>' panicked at crates/analyzer/src/symbol_table.rs:892:
called `Option::unwrap()` on a `None` value
   ...
   veryl_analyzer::symbol_table::SymbolTable::lookup_member
   veryl_analyzer::symbol_table::SymbolTable::resolve
   veryl_analyzer::symbol_table::SymbolTable::get_package
   veryl_analyzer::symbol_table::SymbolTable::apply_import
   veryl_analyzer::analyzer::Analyzer::analyze_post_pass1
   veryl_ls::server::Server::serve
```

The background thread dies, so the progress notification never reaches `end` and `cargo test -p veryl-ls tests::progress` times out.

## Root cause

`Server::background_analyze` clears a file's stale state with `drop_tables` before re-registering it, and `drop_tables` was keyed by `PathId` alone.

When the same path belongs to two projects, analyzing it for the second project dropped everything the first project had registered:

1. `symbol_table::drop` removed the symbols from `symbol_table`, but the scope tree keeps its own `locals` map, which still held those `SymbolId`s. `lookup_member` then hit a dangling id in `self.symbol_table.get(&id).unwrap()`.
2. `scope::drop_tokens` removed the token scopes, while the reference candidates registered by the first project remained. `check_complex_identifier` then hit `scope::token_scope(token.id).unwrap()` returning `None`.
3. `definition_table::drop` removed the definitions, leaving the first project's symbols pointing at `DefinitionId`s that no longer resolve.

`veryl build` / `veryl check` never call `drop_tables`, which is why this only shows up in the language server.

## Fix

Make the drop key `(project, path)` so each project only drops what it registered.

| Table | Change |
| --- | --- |
| `symbol_table::drop` | filter also by `symbol.namespace.paths[0] == prj` |
| `scope::drop_tokens` | filter also by the project of the token's scope (`ScopeArena::project_of`) |
| `definition_table::drop` | entries now carry the project; `insert` / `insert_with_id` take it |
| `text_table` / `attribute_table` / `unsafe_table` | unchanged — keyed by path plus line/column, so they are project independent and re-registering is idempotent |

`SymbolTable::drop` also filters the reference tokens it removes by project, via the new `scope::token_project`.

Call sites updated: `ls/server.rs` (`background_analyze`, `on_change`, `on_remove`), `ls/incremental.rs`, `veryl/incremental.rs`. `fragment_cache::restore` takes the project so that restored definitions are tagged correctly.


crates/languageserver/src/server.rs
```diff
@@ -841,11 +841,12 @@ impl Server {
...
 
-    fn on_remove(&mut self, path: Url) {
-        if let Some(path) = path.to_file_path()
+    fn on_remove(&mut self, url: Url) {
+        if let Some(metadata) = self.get_metadata(&url)
+            && let Some(path) = url.to_file_path()
             && let Some(path_id) = resource_table::get_path_id(path.to_path_buf())
         {
-            drop_tables(path_id);
+            drop_tables(metadata.project.name.as_str().into(), path_id);
```
dalance: By this change, the following case will fail to drop because `project.name` is `veryl_sample` and the registered name is `veryl_sample3`?

```
veryl_sample3 = {version = "0.18.0", project = "veryl_sample", git = "https://github.com/veryl-lang/sample"}
```

Is the project fileter necesarry when `on_remove`?

taichi-ishitani: Is this hook called when the given file was removed?
If yes, filter by project name is not needed.

taichi-ishitani: I chagned the `prj` arg to an optional arg. The filter by project name is ignored if it is set to `None`.

PR #3119, opened 2026-07-28
Introduce `$prop` namespace

### LINKED ISSUE #3111: Namespace of project property
#2968 introduced project property like below,
but the syntax may be confusable because the introduced name by property suddenly appears.

```systemverilog
pub package qux_pkg {
    const QUX_0: i64   = QUX_0_VALUE; // what is QUX_0_VALUE?
}
```

How about adding explicit namespace for property.

```systemverilog
pub package qux_pkg {
    const QUX_0: i64   = $prop::QUX_0_VALUE; // QUX_0_VALUE comes from property
}
```

@taichi-ishitani How about this issue?

taichi-ishitani:
close veryl-lang/veryl#3111

Symbol for `$prob` namespace is inserted directly under the project namespace. Project properties are inserted into the namespace.

### DIFF
--- a/crates/analyzer/src/analyzer.rs
+++ b/crates/analyzer/src/analyzer.rs
@@ -77,12 +77,12 @@ fn insert_namespace_symbol(name: &str, public: bool) -> StrId {
 }
 
 fn insert_project_property_symbols(
-    prject_name: StrId,
+    project_name: StrId,
     properties: &BTreeMap<String, ProjectProperty>,
 ) {
-    let mut namespace = Namespace::new();
-    namespace.push(prject_name);
+    let prop_symbol = create_prop_namespace_symbol(project_name);
 
+    let namespace = prop_symbol.inner_namespace();
     for (prop_name, prop_value) in properties {
         let token = Token::from_external_text(prop_name);
         let value_property = ProjectPropertyValueProperty::new(prop_value, token.into());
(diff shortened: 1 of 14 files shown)


taichi-ishitani:
#3118 fixes CI failure of language sarver.

crates/analyzer/src/analyzer.rs
```diff
@@ -80,6 +80,8 @@ fn insert_project_property_symbols(
     prject_name: StrId,
     properties: &BTreeMap<String, ProjectProperty>,
 ) {
+    symbol_table::add_project_local(prject_name, "$prop".into(), prject_name);
```
dalance: This function will be called from the root project, so `project_local_table` will become to exist for the root.
If so, https://github.com/veryl-lang/veryl/blob/08d7ec62e36e207b8847b830cd04ef8bb9620a05/crates/analyzer/src/symbol_table.rs#L1173 will set `root_prj` to `false` for the root.

crates/analyzer/src/symbol.rs
```diff
@@ -1047,6 +1047,7 @@ pub enum SymbolKind {
     Embed,
     TbComponent(TbComponentProperty),
     ProjectProperty(ProjectPropertyValueProperty),
+    PropNamesapce,
```
dalance: It seems to be typo

taichi-ishitani: I fixed it.

PR #3131, opened 2026-08-02
fix(simulator): support array literal function arguments

tignear:
Expand array literals using the formal function argument type before converting their elements into simulator expressions. Previously, passing an array literal directly to a function caused the simulator to panic with `unhandled Expression variant`.

```veryl
function pick (
    x: input logic<4> [2],
) -> logic<4> {
    return x[1];
}

always_ff (clk) {
    out_q = pick('{in_hi, in_lo});
}
```

The added tests cover unpacked, packed, single-element, repeated, default-filled, and multidimensional literals, including nested function calls.

### DIFF
--- a/crates/analyzer/src/ir/function.rs
+++ b/crates/analyzer/src/ir/function.rs
@@ -1,5 +1,7 @@
 use crate::conv::Context;
-use crate::conv::utils::{check_compatibility, check_implicit_clock_conversion};
+use crate::conv::utils::{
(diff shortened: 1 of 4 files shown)


tignear:
I found that malformed array literal arguments bypass the analyzer. This PR avoids the existing panic, but reports only a generic unsupported description error. I’m converting it back to draft while I fix the diagnostics. Sorry I missed this.

Issue #3134, opened 2026-08-03
Native interpreter silently terminates non-progressing stepped for loops


tignear:
## Summary

The native interpreter silently terminates a dynamically bounded stepped `for` loop when its update does not advance the loop variable. The emitted SystemVerilog instead continues executing the loop indefinitely.

This issue does not depend on integer overflow(#3130).

## Reproducer

```veryl
module StepStall (
    start: input  logic<32>,
    n    : input  logic<32>,
    count: output logic<32>,
) {
    always_comb {
        count = 0;

        for i in start..n step *= 2 {
            count += 1;
        }
    }
}

#[test(step_stall)]
module step_stall {
    var count: logic<32>;

    inst dut: StepStall (
        start: 0,
        n    : 4,
        count: count,
    );

    initial {
        $assert(count == 1);
        $finish();
    }
}
```

The end bound is intentionally dynamic. The equivalent loop with a statically known bound is rejected during analysis.

Tested with Veryl 0.20.3 at commit `b22fb965a983c2d96fc869e41f76439bc3ebc8a1`.

## Observed behavior

| Execution path | Observed behavior |
|---|---|
| Native interpreter | Executes the body once, detects that `0 * 2 == 0`, silently exits the loop, and completes successfully with `count == 1` |
| Emitted SystemVerilog | Continues executing the loop without advancing simulation time because `i` remains zero |

The emitted SystemVerilog is:

```systemverilog
for (int i = start; i < n; i *= 2) begin
    count += 1;
end
```

## Specification note

IEEE 1800-2023 §9.2.2.2 recommends that tools warn when an `always_comb` procedure does not represent combinational logic. However, it does not define a non-progressing procedural loop as a language-level runtime error. A SystemVerilog `for` loop also has no general requirement that its update expression make progress.

Therefore, treating non-progress as an error would be a Veryl-specific language rule or a simulator-specific runtime diagnostic rather than a requirement inherited from SystemVerilog.

Native simulation and emitted SystemVerilog do not necessarily need to behave identically if the Veryl specification permits multiple behaviors. The relevant relationship would instead be:

```text
Beh(native(P))     ⊆ Beh_Veryl(P)
Beh(emitted_sv(P)) ⊆ Beh_Veryl(P)
```

The consequences of possible Veryl semantic definitions can be summarized as follows:

| Possible Veryl semantics | Diagnostic termination by the native simulator | Nontermination of emitted SystemVerilog | Silent successful termination by the native simulator |
[...]

Issue #3130, opened 2026-08-02
Native simulator and generated SystemVerilog disagree on for-loop overflow


tignear:
Veryl emits runtime loop variables as 32-bit signed SystemVerilog int, but the native simulator evaluates the following `*=` and `<<=` updates at a wider width.
```
module MulRepro (
    end_bound: input signed logic<64>,
    hits: output logic<32>,
) {
    always_comb {
        hits = 0;
        for _i in 1500000000..end_bound step *= 2 {
            hits += 1;
        }
    }
}

module ShlRepro (
    end_bound: input signed logic<64>,
    hits: output logic<32>,
) {
    always_comb {
        hits = 0;
        for _i in 1073741824..end_bound step <<= 1 {
            hits += 1;
        }
    }
}

#[test(repro)]
module repro {
    inst clk: $tb::clock_gen;
    inst rst: $tb::reset_gen(clk);

    var mul_hits: logic<32>;
    var shl_hits: logic<32>;

    inst mul: MulRepro (
        end_bound: 64'sd3100000000,
        hits: mul_hits,
    );

    inst shl: ShlRepro (
        end_bound: 64'sd2147483649,
        hits: shl_hits,
    );

    initial {
        rst.assert();
        clk.next(1);
        $assert(mul_hits == 2);
        $assert(shl_hits == 2);
        $finish();
    }
}
```

veryl test passes: the simulated counters reach 3000000000 and 2147483648, respectively.
veryl build emits 32-bit int loop variables. Those values overflow into negative numbers, while both positive 64-bit bounds are greater than INT32_MAX, so the generated SystemVerilog loops do not terminate.
The native simulator and generated SystemVerilog should have consistent counter-width and overflow semantics.

ranaumarnadeem:
@dalance is it ok if i work on this issue?

dalance:
@tignear might already have a patch ready. If not, feel free to work on it.

tignear:
@ranaumarnadeem @dalance 
I haven't checked the SV specifications yet, so I don't have a patch. Even without looking it up, though, my intuition is that the SV spec would likely allow for both errors and infinite loops.

ranaumarnadeem:
@tignear @dalance i'll check the specs and issue and let you guys know or make a PR directly

ranaumarnadeem:
@dalance I spent some time on this.

On the spec question - `int` is 32-bit signed and an overflowing loop variable that never
[...]

PR #3140, opened 2026-08-04
Issue#3130 forloop overflow

ranaumarnadeem:
Refs #3130.

The loop variable is emitted as a 32-bit signed `int` while the simulator
iterates at a wider width, so a step that goes past i32::MAX makes the two
disagree on the trip count.

This rejects those loops at analysis time, same approach as 3319e4f0 for
degenerate steps. The scan is anchored on a constant start so it still works
when the end bound is only known at runtime, which is the case in the report.

Catches both repros from the issue, plus the case where both bounds fit but the
step still carries past the limit. No emitter change, so no golden output moves.

What this doesn't cover:

- A non-constant start, e.g. `for i in start_port..end_port step *= 2` - there's
  nothing to anchor the scan on. @tignear's 1.5e9 -> 1.6e9 example is this shape.
- Runtime end bound with an additive step. A geometric step reaches the limit in
  ~30 iterations, but `+= 1` would need 2^31, so scanning it isn't practical.
- The simulator side. The interpreter keeps the counter in an unmasked 64-bit
  value and the Cranelift backend zero-extends to 32-bit unsigned, so the
  backends still disagree with each other. Happy to do that here or separately.

One thing to flag: as @tignear pointed out, overflow doesn't always mean an
infinite loop - it can wrap past the bound and stop. This check rejects those
cases too, on the grounds that the interpreter and emitted SV still disagree on
the iteration count. Tell me if that's too strict.

#3134 is not really related to this I will see and make a seprate PR for it

### DIFF
--- a/crates/analyzer/src/analyzer_error.rs
+++ b/crates/analyzer/src/analyzer_error.rs
@@ -758,6 +758,21 @@ pub enum AnalyzerError {
 
     #[diagnostic(
         severity(Error),
+        code(for_loop_overflow),
+        help("keep the loop variable within the 32-bit signed range, or narrow the range or step"),
+        url("https://doc.veryl-lang.org/book/07_appendix/02_semantic_error.html#{}", self.code().unwrap())
+    )]
+    #[error("for-loop step overflows the 32-bit signed loop variable")]
+    ForLoopOverflow {
+        #[source_code]
+        input: MultiSources,
+        #[label("Error location")]
+        error_location: SourceSpan,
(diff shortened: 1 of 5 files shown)


dalance:
> One thing to flag: as @tignear pointed out, overflow doesn't always mean an infinite loop - it can wrap past the bound and stop. This check rejects those cases too, on the grounds that the interpreter and emitted SV still disagree on the iteration count. Tell me if that's too strict.

I think it is not too strict, so I'll merge this PR.

Issue #2759, opened 2026-06-08
combinational loop false positive when unrelated assigns share the same comb block


MurakamiShun:
When multiple unrelated assignments exist in the same `always_comb` block, a false positive combinational loop is reported.

In [`build_module_graph`, when `assign_target` is `None`](https://github.com/veryl-lang/veryl/blob/1ad713b9c7c03ce318a6528478f8b4e1a94bb1c8/crates/analyzer/src/comb_loop_detect.rs#L429-L434), the fallback collects all write destinations within the same `reader_decl`. This causes unrelated assignments (e.g. `var3 = var0` and `var4 = var1`) to be incorrectly linked, creating a fake cycle between `var3` and `var4`.

## Veryl code

```
module FPComp (
    op1: input logic<32>,
    op2: input logic<32>,

    less_than: output logic,
) {

    struct FP32 {
        sign: logic    ,
        exp : logic<8> ,
        frac: logic<23>,
    }

    var op1_fp32: FP32;
    var op2_fp32: FP32;

    always_comb {
        op1_fp32 = op1;
        op2_fp32 = op2;

        if (op1_fp32.exp == op2_fp32.exp) {
            less_than = op1_fp32.frac <: op2_fp32.frac;
        } else {
            less_than = op1_fp32.exp <: op2_fp32.exp;
        }
    }
}
```

## Console Output
```
Error: combinational_loop (link)

  × combinational loop detected on "op1_fp32"
    ╭─[/home/shun/Desktop/veryl_test/src/test.veryl:18:9]
 17 │     always_comb {
 18 │         op1_fp32 = op1;
    ·         ────┬───
    ·             ╰── Error location
 19 │         op2_fp32 = op2;
    ╰────
    ╭─[/home/shun/Desktop/veryl_test/src/test.veryl:19:9]
 18 │         op1_fp32 = op1;
 19 │         op2_fp32 = op2;
    ·         ────┬───
    ·             ╰── involved in loop
 20 │ 
    ╰────

```

### Module IR
```
module FPComp {
  input var0(op1): logic<32> = 32'hxxxxxxxx;
  input var1(op2): logic<32> = 32'hxxxxxxxx;
  output var2(less_than): logic = 1'hx;
  var var3(op1_fp32): struct {sign: logic<1>, exp: logic<8>, frac: logic<23>} = 32'hxxxxxxxx;
[...]

PR #3144, opened 2026-08-04
feat(analyzer): rebuild combinational analysis with sparse region MemorySSA

tignear:
Replace the current combinational-loop dependency construction with a control-flow-aware, region-based MemorySSA analysis.

The new analysis preserves:
* procedural order and reaching definitions
* partial and disjoint object regions
* value, control, and address dependencies
* function and module feedthrough
* source provenance for diagnostics
* explicit incomplete-analysis results

This PR also adds the IR-independent `veryl-causal` crate and documents the intended analysis contract in `crates/analyzer/COMBINATIONAL_LOOP_ANALYSIS.md`.

## Motivation

This work has two immediate motivations.

First, false positive caused by the current declaration-level dependency aggregation:

```veryl
always_comb {
    x[0] = a;
    $display("x3=%0d", x[3]);
    o = x[3];
}

assign x[1] = x[0];
assign x[2] = x[1];
assign x[3] = x[2];
```

The actual causal chain is:

```text
a -> x[0] -> x[1] -> x[2] -> x[3] -> o
```

The read of `x[3]` reactivates the procedure and supplies `$display` and `o`, but it does not determine the assignment to `x[0]`.

The current representation loses that distinction and invents an `x[3] -> x[0]` dependency. Restoring the old overlap check would instead miss real cross-bit cycles:

```veryl
assign x[1] = x[2];
assign x[2] = x[1];
```

This is the same underlying class of problem as #2759: once reads and writes are aggregated at declaration level, the analyzer can no longer reliably determine which read affects which write.

Second, #3135 needs a reusable dependency analysis for mutable `for` bounds.

Direct procedural mutation can be detected locally, but indirect influence through combinational logic depends on the same dependency relation used for combinational-loop detection. It also needs to distinguish established stability from an incomplete result when hierarchy, external components, or unsupported effects prevent a complete proof.

Maintaining another dependency collector for that diagnostic would duplicate the same procedural-order, region, function, and hierarchy logic.

## Implementation

Each combinational procedure is lowered to an immutable CFG containing ordered read and write events.

Sparse region MemorySSA resolves reads to their reaching definitions and produces a reusable procedure summary. Branches, loops, early exits, partial writes, and dynamic writes use the same reaching-definition model.

The region domain is partitioned only at observed access boundaries. Analysis cost therefore depends primarily on accesses, definitions, phi nodes, and CFG edges rather than directly on the declared width of a signal or array.

Supported copies, selections, concatenations, and pointwise operations preserve positional dependencies. Operations without a positional model may fall back to conservative whole-operand dependencies without discarding procedural order or unrelated region precision.

Retained entry state from an incomplete assignment is kept separate from an explicit read of the previous value. This prevents incomplete assignments from being mistaken for combinational feedback.

Known function effects and module feedthrough are summarized and composed across call and instance boundaries. FF state remains a combinational boundary.

Observer-only reads such as `$display` arguments remain separate from value dependencies. Sensitivity and value causality are related, but they are not the same graph.

## Parallel procedure analysis

`veryl-causal` owns no parser resources, global symbol-table locks, source locations, or Veryl IR references.

Each combinational procedure and function specialization is therefore an independent analysis work item. CFG, SSA, MemorySSA, and procedure-summary construction can run in parallel.

The resulting summaries are immutable and deterministic and are merged afterward in source or module-topology order.

## Incomplete analysis

The analysis distinguishes:

* a known dependency cycle
* established acyclicity
* incomplete analysis

Without this distinction, absence of a loop diagnostic may mean either that the relevant graph was proved acyclic or that a dependency was silently omitted.

This information is also intended for #3135, where mutable loop-bound analysis needs to distinguish an established error, established stability, and an unknown result.

Opaque components, unresolved hierarchy, `inout`, unsupported effects, and similar boundaries are recorded as incomplete rather than converted into guessed feedthrough for a hard error. A known cycle remains reportable even when unrelated parts of the design are incomplete.

## Why MemorySSA

I considered direct symbolic evaluation, sparse forward dataflow, direct dependence-graph construction, and lowering to a more explicit combinational representation.

Sparse region MemorySSA appeared to be the least complicated option because reaching definitions, control-flow joins, partial writes, diagnostic provenance, incomplete results, and parallel procedure analysis can share one representation.

Another possible implementation would require a specialized pipeline combining symbolic evaluation, range-based stores, expression trees, separate observer tracking, bit-boundary atomization, dependency scheduling, and cycle detection.

MemorySSA is not part of the language contract and is not claimed to be the only possible implementation. It was simply the smallest approach I found which retained the information required by the analyzer.

The implementation is still substantial because the problem covers procedural ordering, regions, control flow, functions, hierarchy, observers, diagnostics, and incomplete behavior—not only SCC detection.

## User-visible effects

The accepted set may change in both directions.

The new analysis can remove false dependencies which caused valid designs to be rejected, but it can also restore real dependencies which the previous implementation omitted.

Diagnostics now use deterministic, source-backed simple cycles.

## Tests

The expanded tests cover:

* the regression exposed by #3033
* the false-positive pattern from #2759
* real and acyclic cross-bit paths
* observer reactivation without false value dependencies
* procedural read-before-write and write-before-read
* dead and dominating writes
* branches, loops, and early exits
* partial and dynamic accesses
* functions and module feedthrough
* FF boundaries
* sparse oversized objects
* deterministic cycle selection
* explicit incomplete reasons

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6123,6 +6123,7 @@ dependencies = [
  "num-bigint",
  "num-traits",
  "paste",
+ "petgraph",
  "postcard",
  "rayon",
  "serde",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6124,6 +6124,7 @@ dependencies = [
  "num-traits",
  "paste",
  "postcard",
+ "rayon",
  "serde",
  "serde_json",
  "similar 3.1.1",
@@ -6134,6 +6135,7 @@ dependencies = [
  "thiserror",
  "toml",
  "vcd",
+ "veryl-causal",
  "veryl-component-sys",
  "veryl-metadata",
  "veryl-parser",
@@ -6151,6 +6153,10 @@ dependencies = [
  "toml",
  "veryl-path",
 ]
+
+[[package]]
+name = "veryl-causal"
+version = "0.20.3"
 
 [[package]]
 name = "veryl-component"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ members = [
     "crates/aligner",
     "crates/analyzer",
     "crates/cache",
+    "crates/causal",
     "crates/component",
     "crates/component/macros",
     "crates/component/sys",
@@ -69,6 +70,7 @@ pprof               = {version = "0.15.0", features = ["flamegraph"]}
 pulldown-cmark      = "0.13.4"
 rand                = {version = "0.10", default-features = false}
 rand_pcg            = "0.10"
+rayon               = "1.12"
 regex               = "1.12.4"
 scnr2               = "0.5.2"
 semver              = {version = "1.0", features = ["serde"]}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,6 +65,7 @@ once_cell           = "1.21"
 parol               = "5.0"
 parol_runtime       = "5.0"
 paste               = "1.0"
+petgraph            = "0.8.3"
 postcard            = {version = "1.1", features = ["alloc"]}
 pprof               = {version = "0.15.0", features = ["flamegraph"]}
 pulldown-cmark      = "0.13.4"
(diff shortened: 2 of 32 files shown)


tignear:
It’s become too large, so I’m splitting it up and submitting a PR.

PR #3161, opened 2026-08-08
Replace combinational-loop analysis with sparse SSA

### LINKED ISSUE #3164: `CombinationalLoop` false positive for ordered reassignment of module-scope variables in `always_comb`
## Overview

Hello, I found a false-positive `CombinationalLoop` diagnostic when module-scope variables are reassigned in procedural order within a single `always_comb` block. This issue is reproducible with Veryl 0.20.3-nightly (9e585f2 2026-08-09).

## Reproduction

```
module top (
    a: output logic,
) {
    var b: logic;

    always_comb {
        a = 0;
        b = a;
        a = b;
    }
}
```

Running `veryl check` reports:

```
Error: combinational_loop
× combinational loop detected on "a"
```

The diagnostic involves all three assignments.

## Expected behavior

This block is feed-forward under Veryl's blocking assignment semantics for `always_comb`.

The first assignment defines `a` before it is read. Distinguishing successive definitions, the dependency chain is:

0 -> $a_1$ -> $b_1$ -> $a_2$

The read of `a` in `b = a` reaches the preceding `a = 0`, and the final `a = b` reads the newly assigned `b`. Neither read depends on the value of `a` or `b` from before the block executes.

Therefore, this code should not be diagnosed as a combinational loop.

tignear:
Replace the existing combinational-loop detector with statement-ordered SSA and sparse bit/array partitions.
This removes the array-size cutoff without enumerating the full declared shape and preserves element identity for large and multidimensional arrays.
This PR provides the initial implementation only. Remaining false positives and false negatives, richer function/interface summaries, and latch diagnostics will be improved incrementally in follow-up PRs.
`veryl build` was run against every Veryl project collected in `support/discovery`, including `pzbcm`.

fix #3164


dalance:
I think this PR can be merged, but I found some problematic pattens.
Could you check them?

* false-positive pattern

```
module Top (
    p: output logic,
    q: output logic,
) {
    function f (
        x: input  logic,
        y: output logic,
    ) {
        if x {
            y = 1;
        }
    }
    var a: logic;
    var b: logic;
    always_comb {
        f(a, p);
        f(b, q);
    }
    assign a = q;
    assign b = 0;
}
```

* O(N^2) performance

|N|master|this PR|
|-|-|-|
|1024|0.19|0.21|
|8192|0.63|3.11|
|16384|1.61|13.99|

```
var mem: logic<32> [N];
var acc: logic<32> [N];
always_comb {
    sum = 0;
    for k in 0..N {
        acc[k] = mem[k] ^ i_d;
        sum    = sum + acc[k];
    }
}
```

tignear:
<html>
<body>
<html><head></head><body><p><span>Thank you for the concrete examples.</span></p><p><span>The false positive was caused by function-local SSA bindings being shared between function invocations. The O(N²) behavior was caused by repeatedly scanning all sparse array partitions during both partition construction and lookup.</span></p><p><span>I fixed them by isolating SSA state per invocation, constructing partitions with an event sweep, and indexing intervals per variable.</span></p><p>

Post-fix measurements (median of seven local runs):

| N     | Time (s) |
|------:|---------:|
| 1,024 | 0.531    |
| 8,192 | 1.397    |
| 16,384 | 2.621   |

<p><span>These are medians of seven local runs. I also removed the input-size-dependent recursive traversals found while validating this case.</span></p><p><span>Thank you for catching these issues.</span></p></body></html>
</body>
</html>

dalance:
Thanks!
I'll merge this PR.

PR #3169, opened 2026-08-10
fix std repository path in Veryl.toml

Kenta11:
On https://std.veryl-lang.org/, the repository link points to an archived repository. It was not updated when std was merged into the veryl repository. It might be better to point the link at the current location so readers are not sent to an archived repository.

### DIFF
--- a/crates/std/veryl/Veryl.toml
+++ b/crates/std/veryl/Veryl.toml
@@ -4,7 +4,7 @@ version     = "0.0.1"
 authors     = ["dalance@gmail.com"]
 description = "Veryl Standard Library"
 license     = "MIT OR Apache-2.0"
-repository  = "https://github.com/veryl-lang/std"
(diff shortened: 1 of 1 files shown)


crates/std/veryl/Veryl.toml
```diff
@@ -4,7 +4,7 @@ version     = "0.0.1"
 authors     = ["dalance@gmail.com"]
 description = "Veryl Standard Library"
 license     = "MIT OR Apache-2.0"
-repository  = "https://github.com/veryl-lang/std"
+repository  = "https://github.com/veryl-lang/veryl/tree/master/crates/std/veryl"
```
dalance, line 7: `repository` field should be repository URL which doesn't include inner path.
So `veryl-lang/veryl` is correct.

Kenta11: I see. The `repository` points to a git repository URL. Closing this pr.

PR #3176, opened 2026-08-10
Extend combinational loop analysis to function effects and instance receivers

tignear:
parent: #3161 
Extend combinational-loop analysis to track function effects in instance inputs and preserve interface receiver identity across instances.

Also reuse the module analysis context across always_comb declarations. For the same generated 1,024-declaration module, this reduced `comb_loop_detect::check` time from 6.20 s before context reuse to 0.24 s after it in a local debug build.


dalance:
I found two edge cases.
Could you check them?

```systemverilog
interface Bus {
    var value: logic[2];
    function get0 () -> logic { return value[0]; }
    function put0 (next: input logic) { value[0] = next; }
}
module Top {
    inst bus: Bus;
    always_comb {
        bus.put0(bus.get0());   // not detected
        bus.put1(0);
    }
}
```

```systemverilog
module Sink (i: input logic) {}
module Top {
    var q: logic;
    var p: logic;
    function f (a: input logic) -> logic {
        p = q;
        q = a;
        return 0;
    }
    inst u: Sink (i: f(0));   // not detected
    always_comb { q = p; }
}
```

tignear:
Thanks — both are genuine false negatives.

- The captured `p`/`q` writes in the instance actual are fixed and the regression is enabled in #3242.
- Assuming `put1` writes `value[1]`, the `put0(get0())` case followed by that disjoint write is fixed and enabled in #3243.

These fixes depend on the positional SSA groundwork in #3241, so pulling them into #3176 would also pull in the later analysis layers. I opened the follow-ups as drafts to make the dependency visible. Sorry for the number of PRs; would it be okay to keep these regressions ignored in #3176 and enable them in the PRs above?

dalance:
OK.
I am currently in the process of merging a sequence of some PRs, so I will merge this one once that is finished.

PR #3178, opened 2026-08-10
fix doc HTML escaping

Kenta11:
When veryl doc parses a veryl file that contains <> in its param, it outputs <> as an HTML tag.

[Example:](https://std.veryl-lang.org/counter.html)

image

Though MAX_COUNT, MIN_COUNT, INITIAL_COUNT, and WRAP_AROUND are params, they are shown under Ports. The HTML file is broken.  "<" and ">" must be escaped.


dalance:
Looks good!
I'll merge this PR.

PR #3185, opened 2026-08-10
diag(simulator): per-step FF digest, ranged state dumps, and offset-to-name lookups

dalance:


dalance:
Dropped: bisect probes are kept as an on-demand patch instead of upstream code (a probe session requires a rebuild either way).

Issue #1588, opened 2025-05-04
[Feature] Use components defined in other projects without project prefix


taichi-ishitani:
We need to add the project prefix to the component name if it is defined in the other project like this.

`$std::axi3_if::<$std::axi3_pkg::<32, 4, 8>>`

I think it is useful to use such components without a project prefix, since the prefix can appear multiple times.
We need to add `namespace` import for this purpose, it is similar to `package` import.

```systemverilog
import $std::axi3_if, $std::axi3_pkg;
inst axi_if: axi3_if::<axi3_pkg::<32, 4, 8>>;
```

PR #3198, opened 2026-08-12
feat(analyzer): allow importing component namespaces (package, module, interface)

### LINKED ISSUE #3122: Consider allowing package namespaces in import declarations
Hello,

I noticed that importing a package itself is currently rejected:

```veryl
import fixedpoint::Q8_24;

module Example(audio: output Q8_24::Raw) {
    assign audio = Q8_24::ZERO;
}
```

This produces `invalid_import`. The current analyzer tests also explicitly include `import PkgA;` as an invalid import, so I understand that this may be intentional.

However, the documentation does not explain that package namespaces cannot be imported directly, and the error message does not indicate why the target is invalid.

Would it be reasonable to allow this form? It would make the origin of a type clear at the use site:

```veryl
Q8_24::Raw
Q1_31::Raw // different fixed-point format's "Raw"
```

Currently, `import Q8_24::*;` and `import Q8_24::Raw;` can be used, while `alias package` is another option. However, an alias seems unnecessary when no renaming is intended. Importing the package namespace would also be closer to the way module namespaces are commonly used in Rust.

If this direction seems useful, I can also try implementing it. If it is intentionally unsupported, a more specific diagnostic and an explanation in the import documentation would be helpful.

AkiyukiOkayasu:
Fixes #3122
Related: #1588

## Description

Importing a package itself was rejected with `invalid_import` even though the import path resolves to a package. 
This change allows a component (package, module or interface) to be imported under its own name, so its members can be referenced with the component name as a qualifier:

```veryl
import dep::MyPkg;
import dep::MyIf;

module Example(audio: output MyPkg::Raw, ifp: modport MyIf::mp) {
    assign audio = MyPkg::ZERO;
    inst u: MyIf;
}
```

## Changes

1. Extend the import checker (conv/checker/import.rs)
   - Allow modules and interfaces as non-wildcard import targets in addition to packages (`is_package` -> `is_component`)

2. Import generic components as-is (reference_table.rs)
   - A generic package/module/interface without generic arguments is imported as-is so it can be instantiated at the use site (`import dep::MyPkg;` then `MyPkg::<16>::Raw`); wildcard imports of instantiated packages (`import pkg::<32>::*;`) keep their existing behavior
   - Restrict the as-is rule to the final path segment so `import pkg::<32>::alias;`-style paths still resolve

3. Fix generic argument resolution (symbol_table.rs)
   - Decide the project prefix of generic arguments from the base component's declaration scope instead of the resolution cursor
   - A head reached through an import binding leaves the cursor in the importing project, which suppressed the prefix and left a same-project struct passed as a generic argument of an imported dependency component unresolvable during member access

4. Tests
   - Update the `invalid_import` unit test (package/module/interface imports are now valid)
   - Add testcases for file-scope, module-scope, and generic component namespace imports (25_dependency_2)

## Check

- `cargo fmt --check`
- `cargo clippy -p veryl-analyzer --all-targets -- -D warnings`
- `cargo test`
- `make gen_sv` + `git diff --exit-code`
- std: `veryl fmt --check` / `veryl check` / `veryl test --sim verilator`


taichi-ishitani:
How about supporting other components such as modules and interfaces?
https://github.com/veryl-lang/veryl/issues/1588

AkiyukiOkayasu:
Sounds good, I'll add support for that too.

AkiyukiOkayasu:
@taichi-ishitani I've finished implementing it, so could you please review it?

testcases/sv/25_dependency_2.sv
```diff
@@ -3,6 +3,8 @@
 
 
 module veryl_testcase_Module25A
+    import veryl_sample4_qux_pkg;
```
dalance: Maybe, importing package is not supported by SV, and it is not necessary because imported items are emitted as full path.

AkiyukiOkayasu: I have fixed it. 
The emitter now suppresses the import of single components (packages, modules, interfaces, or instances thereof). SystemVerilog `import` only accepts `package::*` or `package::item`, and references through them are already output as fully qualified paths. 
`should_skip_import` now also covers these types, and a regression test has been added. The generated SystemVerilog no longer includes lines in the format `import veryl_sample4_qux_pkg;`.

crates/analyzer/src/reference_table.rs
```diff
@@ -322,15 +322,26 @@ impl ReferenceTable {
...
                         && !params.is_empty()
                         && (matches!(
                             symbol.found.kind,
                             SymbolKind::Struct(_) | SymbolKind::Union(_)
-                        ) || matches!(symbol.found.kind, SymbolKind::Function(ref x) if !x.is_proto))
+                        ) || matches!(
+                            symbol.found.kind,
+                            SymbolKind::Package(_)
+                                | SymbolKind::Module(_)
+                                | SymbolKind::Interface(_)
```
taichi-ishitani, line 334: For module, interface, and package, generic definitions should be imported but not generic instances. Threfore, I think `n_args == 0` should be added like function import.

taichi-ishitani (changes requested):

crates/analyzer/src/reference_table.rs
taichi-ishitani: Tests showing rejecting generic instnaces for them should also be added.

taichi-ishitani: Sorry, in this context, `n_args == 0` is not necessray. But I'd like you to add those tests.

AkiyukiOkayasu: Thank you for the review. I have added a feature to reject the import of generic component instances (`import pkg::<8>;` now causes an `invalid_import` error for packages, modules, and interfaces; only definitions can be imported and instantiated at use sites). 
I have also added unit tests covering this rejection case as well as cases that remain valid (wildcard instance imports and member imports via instances).

AkiyukiOkayasu:
Let me know if you'd like me to squash the commits. I've left them split up for now to make the review easier.

taichi-ishitani (approved):

dalance:
Looks good.
I'll merge this PR.

PR #3208, opened 2026-08-12
fix(emitter): keep #[ifdef] guards on generated modport members

### LINKED ISSUE #3191: `..converse` drops `#[ifdef]` guards from modport members
## Overview

Hello, I found that `..converse(...)` does not preserve `#[ifdef(...)]` guards on source modport members. As a result, the generated SystemVerilog may contain a modport item referring to a signal removed by the preprocessor. This issue is reproducible with Veryl 0.20.3-nightly (7cd7de6 2026-08-11).

## Reproduction

```veryl
interface TestIf {
    #[ifdef(LOG)]
    var trace: logic;

    var dummy: logic;

    modport master {
        #[ifdef(LOG)]
        trace: input,
        dummy: input,
    }

    modport slave {
        ..converse(master)
    }
}

module Top (
    bus: modport TestIf::slave,
) {}
```

`veryl build` generates the following complete SystemVerilog:

```systemverilog
interface veryl_converse_ifdef_issue_TestIf;
    `endif

    logic dummy;

    modport master (
        `ifdef LOG
        input trace,
        `endif
        input dummy
    );

    modport slave (
        output trace,
        output dummy
    );
endinterface

module veryl_converse_ifdef_issue_Top (
    veryl_converse_ifdef_issue_TestIf.slave bus
);
endmodule
//# sourceMappingURL=repro.sv.map
```

Without `LOG`, the `slave` modport refers to the removed `trace` signal:

```console
$ verilator --lint-only target/repro.sv
%Error: target/repro.sv:16:16: Modport item not found: 'trace'
    16 |         output trace,
    |                   ^~~~~
%Error: Exiting due to 1 error(s)
```

The member generated by `..converse(master)` should preserve the source member's guard:

```systemverilog
modport slave (
    `ifdef LOG
    output trace,
    `endif
    output dummy
);
```

ranaumarnadeem:
Closes #3191.

`..converse(...)` (and `..same`/`..input`) generated modport members without their
`#[ifdef]` guards, so the emitted SV referenced signals that don't exist when the
define is off:

```systemverilog
modport slave (
    output trace,     // no `ifdef LOG
    output dummy
);
```

The cause is that these members don't exist in the source - they're generated with
`Token::generate`, which has no line/column. Attributes are stored in a positional
table keyed by token range, so there's nothing to look up for a generated token
(trying to insert one actually hits the `unreachable!()` in `range_table.rs`). The
guards therefore have to come from the variable or function the member was
generated from, which does have a real position. That's what this does.

The fiddly part is the comma. The existing scheme is "trailing comma inside the
item's own guard", which works for a guarded item that isn't last, and
`last_item_with_define` rejects the guarded-last case because it would leave a
dangling comma. That check doesn't cover generated members though - their order
follows the interface declaration order, so a guarded member can end up last:

```veryl
var dummy: logic;
#[ifdef(LOG)]
var trace: logic;
```

So instead of a fixed position, the comma is owned by whichever side is guarded: a
guarded member keeps it inside its own guard, so the comma disappears along with the
member. That means a guarded member at the end emits a leading comma, which looks a
bit unusual but is valid either way:

```systemverilog
modport slave (
    output dummy
    `ifdef LOG
    , output trace
    `endif
);
```

Guards are taken from the variable rather than the source modport member on purpose:
if a signal is conditional, any modport referring to it has to be guarded regardless
of how the origin member happened to be annotated.


testcases/sv/75_modport_default.sv
```diff
@@ -70,41 +83,59 @@ interface veryl_testcase_Interface75;
...
+        , input  f
+        `endif
     );
 
     modport partial_same (
         output a    ,
         input  b    ,
         output c    ,
-        output d    ,
+        output d
```
dalance: This is incorrect because if `TRACE` is not defined the generated code becomes `output d import get_a,`.

crates/emitter/src/emitter.rs
```diff
@@ -2732,35 +2732,93 @@ impl Emitter {
...
+                            _ => None,
+                        };
+                        let defines: Vec<_> = source
+                            .and_then(symbol_table::get)
+                            .map(|x| attribute_table::get(&x.token))
+                            .unwrap_or_default()
+                            .into_iter()
+                            .filter_map(|x| match x {
+                                Attr::Ifdef(x) => Some(("ifdef", x)),
+                                Attr::Ifndef(x) => Some(("ifndef", x)),
```
dalance: Are `else` / `elsif` patterns like below not supported?

```
#[ifdef(LOG)]
var trace_a: logic;
#[else]
var trace_b: logic;
```

ranaumarnadeem: Fixed now. They can't be emitted as `else` directly, since a generated member has
no guarded neighbour to chain onto, so each branch is rebuilt as a standalone guard
from the conditions it needs:
```systemverilog
modport slave (
    `ifdef LOG
    output trace_a,
    `endif
    `ifndef LOG
    output trace_b,
    `endif
    output dummy
);
```
elsif chains nest the same way - the elsif branch becomes `ifndef A + `ifdef B.

testcases/sv/75_modport_default.sv
ranaumarnadeem: Fixed. Each member now keeps its own comma inside its own guard, so the comma
disappears along with the member:
```systemverilog
modport master (
    `ifdef TRACE
    input  f    ,
    `endif
    `ifndef TRACE
    input  g    ,
    `endif
    input  a    ,
    input  b    ,
    output c    ,
    output d    ,
    import get_a,
    ...
    import set_d
);
```
The one comma that can't work that way is the last member's, since that's the one
that isn't there - so generated members are now emitted with the guarded ones first
and an unconditional member last. That changes their order, but item order isn't
significant in SV. Say if you'd rather I did it another way.

dalance: For module ports, the issue of trailing commas is avoided by prohibiting the use of `ifdef` on the final element. Wouldn't it be simpler to introduce a similar constraint here, rather than resorting to reordering?

PR #3229, opened 2026-08-16
Detect side-effecting function calls in always_ff

### LINKED ISSUE #3165: Consider assignment semantics for module-scope variables in functions called from `always_ff`
## Overview

Hello, I encountered an interaction between `always_ff` assignment semantics and functions that modify module-scope variables.

It seems that the assignment semantics for this case may not have been explicitly considered or defined. The current generated SystemVerilog can contain both blocking and non-blocking assignments to the same variable, so it may be worth clarifying or handling this case.

This behavior was confirmed with Veryl 0.20.3-nightly (9e585f2 2026-08-09).

## Reproduction

```
module top (
    clk: input clock,
    q  : output logic<2>,
    r  : output logic<2>,
) {
    function update () {
        q = 1;
    }

    always_ff {
        q = 2;
        update();
        r = q;
    }
}
```

veryl build succeeds and generates:

```v
function automatic void update();
    q = 1;
endfunction

always_ff @(posedge clk) begin
    q <= 2;
    update();
    r <= q;
end
```

The direct assignment to `q` is emitted as a non-blocking assignment, while the assignment in `update()` is emitted as a blocking assignment.

As a result, Verilator reports:

```
%Warning-BLKSEQ: Blocking assignment '=' in sequential logic process
%Error-BLKANDNBLK: Unsupported: Blocked and non-blocking assignments to same variable: 'q'
```

## Question

I may be misunderstanding the intended semantics of this case.

According to the assignment semantics (https://doc.veryl-lang.org/book/05_language_reference/18_execution_model/03_assignment_semantics.html),
assignments to module-level variables in `always_ff` use non-blocking semantics. However, I am not sure whether this is also intended to apply to assignments performed indirectly through a function call.

Currently, the direct assignment to `q` and the assignment inside `update()` are lowered differently, resulting in mixed blocking and non-blocking assignments.

One possible interpretation might be to apply the caller's `always_ff` semantics to the function's assignment. 

Alternatively, if assignments inside functions are intentionally always blocking, perhaps a diagnostic for this combination could be helpful.

tignear:
This change prevents function calls from hiding writes from the `always_ff` scheduler. It transitively tracks external writes and written outputs, rejecting external writes and requiring written outputs to target process-local variables.

Full interprocedural scheduler analysis could identify rare cases where a hidden blocking write is unobservable. In most useful cases, however, the written value is consumed by other scheduled operations, making the difference from non-blocking assignment observable depending on execution order. Complete dependency analysis would therefore permit very few additional cases, so a boolean effect summary provides nearly the same practical boundary.

`$display` and assertions are excluded because they do not modify scheduler-visible HDL state. Conditional compilation, concatenation assignments, and generic/proto functions are covered.

Closes #3165.


dalance:
I agree this design for function semantics of Veryl.
I found an issue about mis-detection of unassigned.
Could you check it?

```systemverilog
module t (i: input logic, o: output logic) {
    #[allow(unassign_variable)]
    function f (y: output logic) {
        $sv::g(y);
    }
    always_comb {
        f(o);
    }
}
```

PR #3241, opened 2026-08-18
Track positional control flow in combinational loop analysis

tignear:
parent: #3176
Track packed and array positions through SSA, dynamic writes, early exits, specialization, and instance boundaries. Incomplete instance selectors remain conservative.


dalance:
A new `AnalyzerError` (`CombinationalLoopPositionOverflow`) seems to be added.
So could you add snapshot test for it?

crates/analyzer/src/comb_loop_detect.rs
```diff
@@ -1428,25 +2105,44 @@ fn check_graph(module: &Module, graph: &Graph<NodeKey, ()>, errors: &mut Vec<Ana
     }
 }
 
-fn strongly_connected_components(graph: &Graph<NodeKey, ()>) -> Vec<Vec<NodeIndex>> {
-    // Petgraph's Tarjan implementation uses recursive DFS. A long, otherwise
-    // shallow dependency chain can therefore exhaust the native stack. The
-    // Kosaraju implementation uses explicit worklists for both passes.
```
dalance, line 1590: Previously Tarjan seems to be rejected because of stack overflow.
So why did you revert to Tarjan?

tignear: I didn't have any particular intention, nor do I recall doing any additional work on this, but I remember clicking the "update with rebase" button, so I'm surprised by how Git behaved. I apologize.

PR #3242, opened 2026-08-18
Analyze runtime control flow and constrained dependency cycles

tignear:
Improves combinational-loop analysis by tracking runtime control flow, including loop iterations, `break`/`return`, and conditional function side effects. Dependency tracking preserves branch exclusivity and bit/array positions rather than merging them into unconditional whole-variable dependencies.

### Detect feedback through control flow

```veryl
module Feedback (o: output logic) {
    var stop: logic;
    assign stop = o;

    always_comb {
        o = 0;
        for _index in 0..1 {
            if stop {
                break;
            }
            o = 1;
        }
    }
}
```

The `break` condition determines the final value of `o`, while `stop` depends on `o`. This is diagnosed as a combinational loop.

### Avoid false positives for shifted dependencies

```veryl
module ShiftChain (o: output logic<16>) {
    var shifted: logic<16>;
    assign shifted[12:0] = o[15:3];
    assign shifted[15:13] = 0;
    assign o = shifted;
}
```

Although the variable-level graph contains `o -> shifted -> o`, the bit-level dependencies form a finite chain, not a cycle. This is not diagnosed as a combinational loop.

parent: #3241


dalance:
The following code seems to be caused infinite loop.
Could you check it?

```
module ShiftDangle (i: input logic<16>, o: output logic<16>) {
    var s: logic<16>;
    var t: logic<16>;
    assign t[12:0]  = s[15:3];
    assign t[15:13] = i[2:0];
    assign s        = t;
    assign o        = i;
}
```

tignear:
> The following code seems to be caused infinite loop. Could you check it?
> 
> ```
> module ShiftDangle (i: input logic<16>, o: output logic<16>) {
>     var s: logic<16>;
>     var t: logic<16>;
>     assign t[12:0]  = s[15:3];
>     assign t[15:13] = i[2:0];
>     assign s        = t;
>     assign o        = i;
> }
> ```

Thank you for reviewing this so carefully. My review lacked consideration regarding the behavior in cases where the output remains unaffected. I will double-check this thoroughly.

PR #3262, opened 2026-08-20
feat(analyzer): allow importing project-scope functions

### LINKED ISSUE #3260: Allow importing project-scope functions
I would like to be able to import functions defined at the top level (project scope) of a dependency project.

## Background

Project-scope functions cannot be imported, and callers are currently required to write them out using fully qualified paths (`dep::func::<W>(...)`) every time.

```veryl
// Currently results in invalid_import
import dep::project_generic;
```

Project-scope functions cannot be imported regardless of whether they have generics or not (non-generic `project_plain` results in the same error). While this can be avoided because functions inside packages can be imported, doing so hits a different limitation (#3110) where module parameters cannot be passed as generic arguments.

## Related

- #3110 (Issue where module parameters cannot be passed as generic arguments, closed)
- #1619/#1624 (As-is import of generic functions, fixed)
- #3198 (PR that added namespace imports)

AkiyukiOkayasu:
Fixes #3260

## Description

Project-scope functions of a dependency can now be imported by their project-qualified path (`import dep::func;`), so call sites can use `func::<...>(...)` instead of repeating the full path.

```veryl
// Dependent project: dep
function project_generic::<W: u32> (value: input logic<W>) -> logic<W> {
    return value;
}

// Caller
import dep::project_generic; // previously invalid_import
let r: logic<8> = project_generic::<8>(8'd1);
```

## Changes

1. Extend the import checker (`conv/checker/import.rs`)

   - Accept the project namespace as a qualifier for import targets. Only functions are accepted there, because the emitted SystemVerilog has no package to qualify a project-scope import with.

2. Emitter (`emitter.rs`)

   - Suppress the bare project-scope function import, which would otherwise produce invalid SystemVerilog (`import dep_func;` without a package); the function is already in scope at the use site.
   - Emit references to an imported global function from another project with the project prefix (`dep_foo_func_2`) so the call matches the emitted definition. This fixes a regression where a non-generic dependency function was called under its bare imported name while its definition was emitted with the project prefix.

3. Tests

   - Add analyzer testcases for generic and non-generic project-scope function imports.
   - Add an emitter test that the bare import is suppressed.
   - Add an integration testcase (Module25T) for a non-generic dependency function import.


dalance:
I found miscompilation caused by this PR.
Could you check it?

```systemverilog
function global_func() -> u32 {
    return 2;
}
module M {
    var _a: u32;
    assign _a = global_func();
}
```

```systemverilog
module prj_M;
    int unsigned _a;
    always_comb _a = prj_global_func();

    function automatic int unsigned global_func() ; // `prj_` is not emitted
        return 2;
    endfunction
endmodule
```

AkiyukiOkayasu:
@dalance Thank you for pointing this out. I was able to reproduce the issue.

The prefix decision was based on whether the global function belonged to a project other than the root project, rather than whether it belonged to the namespace currently being emitted. As a result, calls within a dependency incorrectly received a project prefix, while the copied function definition did not.

I changed the prefix decision to use the destination namespace. I also added regression tests covering:

- A dependency calling its own global function
- A dependency function copied into a root module and calling another global function

Could you please take another look?

dalance:
I think this PR can be merged, so could you resolve the conflict?

AkiyukiOkayasu:
@dalance Thank you for the review. I've resolved the conflicts.

PR #3267, opened 2026-08-21
fix(analyzer): reject a variable, port or const used as a type

Kenta11:
A variable, port or const is accepted where a type is expected: in a variable, constant or function definition, and as a cast target. It must be a bug.

```veryl
module ModuleA (
    i_a: input logic<8>,
) {
    const C: u32 = 8;
    var a: logic<8>;

    var b: i_a;               // emitted as `i_a b;`
    var c: C;                 // emitted as `C c;`
    const D: a = 0;           // emitted as `localparam a D = 0;`
    function g () -> a {      // emitted as `function automatic a g();`
        return 0;
    }
    let z: logic<8> = a as a; // emitted as `a'(a)`
}
```


dalance:
This change seems to reject the following code.
Could you check it?

```systemverilog
module Top {
    const L: logic<32> = 4;
    var a: logic<8>;
    assign a = 0;

    for i in 1..4 :g {
        let b: logic<8> = a as i;   // genvar
    }
    let e: logic<8> = a as L;       // const
}
```

Kenta11:
Thank you for your comment. I fixed it. I also added the genvar and constant cases to `testcases`.

PR #3271, opened 2026-08-22
Resolve generic arguments in their own namespace

### LINKED ISSUE #3163: `$std::axi4_lite_pkg` doesn't have correct parameters when localparams are set to its generics parameter
When I set constant local parameters to the generics parameter slot of `$std::axi4_lite_pkg`, the generated SV doesn't have correct width. The generated code may raise syntax error. If this usage is not permitted, stopping the transpiler should be better.

Input
```
package pkg {
        const ADDR_W: u32 = 64;
        const DATA_BYTES: u32 = 4;
        const ID_W: u32 = 8;

        alias package axi4l_pkg = $std::axi4_lite_pkg::<ADDR_W, DATA_BYTES, ID_W>;
}

module mdl (
        axi4l_m: modport $std::axi4_lite_if::<pkg::axi4l_pkg>::master,
) {
}
```

Output (generated target)
```systemverilog
package hello_pkg;
    localparam int unsigned ADDR_W     = 64;
    localparam int unsigned DATA_BYTES = 4;
    localparam int unsigned ID_W       = 8;


endpackage

module hello_mdl (
    __std___axi4_lite_if____std___axi4_lite_pkg__ADDR_W__DATA_BYTES__ID_W.master axi4l_m
);
endmodule
```

Output (generated interface): multiple signals don't have type declarations.
```systemverilog
interface __std___axi4_lite_if____std___axi4_lite_pkg__ADDR_W__DATA_BYTES__ID_W;
    import ::*;

    //Write address channel
    logic awvalid;
    logic      awready;
     awaddr ;
     awprot ;
     awid   ;
...
```

Output (generated pakcage): local parameters aren't set.
```systemverilog
///### AXI4-Stream bus package
package __std___axi4_lite_pkg__pkg_ADDR_W__pkg_DATA_BYTES__pkg_ID_W;
    localparam int unsigned ADDRESS_WIDTH    = ;
    localparam int unsigned DATA_WIDTH_BYTES = ;
    localparam int unsigned ID_LENGTH        = ;

    typedef logic [ADDRESS_WIDTH-1:0]          addr_t;
    typedef logic [(DATA_WIDTH_BYTES * 8)-1:0] data_t;
    typedef logic [DATA_WIDTH_BYTES-1:0]       strb_t;

    typedef logic [ID_LENGTH-1:0] id_t;

    typedef __std_axi4_lite_config::axprot_bits   proto_t;
    typedef __std_axi4_lite_config::resp_variants resp_t ;

endpackage
```

(I'm not sure why the package is decorated as `AXI4-Stream bus package`...)

Environment
```
$ veryl --version
veryl 0.20.3 (633a884 2026-08-03)
$ uname -a
Linux DESKTOP-RSD9PB4 6.18.33.2-microsoft-standard-WSL2 #1 SMP PREEMPT_DYNAMIC Thu Jun 18 21:54:43 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux
```

taichi-ishitani:
fix veryl-lang/veryl#3163

`68_std_3.veryl` / `68_std_4.veryl` are the reported case:

```veryl
package Package68I {
    const ADDR_W    : u32 = 16;
    const DATA_BYTES: u32 = 4;
    const ID_W      : u32 = 8;

    alias package axi4lite_pkg = $std::axi4_lite_pkg::<ADDR_W, DATA_BYTES, ID_W>;
}

module Module68J (
    axi4lite_m: modport $std::axi4_lite_if::<Package68I::axi4lite_pkg>::master,
) { ... }
```

An alias target is written in the alias declaration's namespace. `unalias` lifts it
out to the referring site but leaves it spelled as it was, so `ADDR_W`, `DATA_BYTES`
and `ID_W` keep a spelling that only resolves inside `Package68I`:

```systemverilog
    __std___axi4_lite_if____std___axi4_lite_pkg__ADDR_W__DATA_BYTES__ID_W.master axi4lite_m
                                              // ^^^^^^ unqualified
        axi4lite_m.awprot  = '(0);
                          // ^ no package qualifier
```

The argument, which `unalias` had just rewritten from `Package68I::axi4lite_pkg` to
`$std::axi4_lite_pkg::<ADDR_W, DATA_BYTES, ID_W>`, reached `append_namespace_path`
with the *referring* namespace, where `ADDR_W` does not resolve, so no namespace
could be appended.

`unalias_inner` now qualifies the target itself, from the alias declaration's
namespace. It has to run *after* the recursive expansion, because
`append_namespace_path` also descends into generic arguments: an argument that is
itself an alias would be qualified before it is expanded, and the added prefix then
keeps the alias from resolving at all.

`append_namespace_path` has a separate defect that shows up once that call is made.
When it prepends the package namespace, the project-prefix decision that follows
still looks at the original head symbol, and `ADDR_W` is a const rather than a
component, so no prefix is added and `Package68I::ADDR_W` stays unresolvable from
`$std` — which is what left the generated package with empty `localparam`s. A head
the call just package-qualified now counts as one.

## A dependency across projects was sometimes not recorded

`type_dag` resolved a generic argument from the base template's namespace, and a
dependency cannot see its consumer:

```
[DAG-ARG]     sym=axi4_lite_pkg sym_ns=[$std] arg=Package68I::ADDR_W
[DAG-RESOLVE] seg=Package68I::ADDR_W ns=[$std] -> None
```

`insert_path` skips a segment it cannot resolve, so `axi4_lite_pkg -> Package68I`
went unrecorded, while the same edge inside a single project was recorded. It now
resolves from the scope the argument's head token was written in. A global function
keeps the template namespace, since it is emitted into the caller and the edge would
be a self-reference against the calling component.

## An alias declaration inserted a dependency it should not

`insert_path` bubbles a reference up to its enclosing component, so
`alias package axi4lite_pkg` became `Package68I -> $std::axi4_lite_pkg` — the reverse
of the real dependency, and enough to close a cycle with it. An alias emits nothing —
it is substituted by its target at the use site — so it carries no dependency of its
own, and its enclosing component has no ordering constraint on the target. Alias
parents are excluded from the graph now.

An alias whose target is parameterized by consts of the declaring package is
therefore no longer a cyclic dependency, so that case is dropped from
`tests::cyclic_type_dependency`. `testcases/error/unresolvable_generic_argument.veryl`
uses `alias module` instead of `inst` for the mirror reason: with the argument now
resolvable, the `inst` spelling also reports a cycle, which is not what that case is
testing.


dalance:
I found two issues about this PR.
Could you check them?

* unexpected cyclic dependency

```systemverilog
module Top {
    const X: u32 = 1;
    inst u: Sub::<X>;
}
module Sub::<X: u32> { let y: logic<X> = 0; }
```

*  #3163 is still panic

```
package pkg {
        const ADDR_W: u32 = 64;
        const DATA_BYTES: u32 = 4;
        const ID_W: u32 = 8;

        alias package axi4l_pkg = $std::axi4_lite_pkg::<ADDR_W, DATA_BYTES, ID_W>;
}

module mdl (
        axi4l_m: modport $std::axi4_lite_if::<pkg::axi4l_pkg>::master,
) {
}
```

taichi-ishitani:
> unexpected cyclic dependency

`module Top` depends on `module sub` and `module sub` depends on `const Top::X` (= `module Top`) so reporting cyclic dependency error is expected behavior.

> panic

a.veryl: Veryl code shown in the issue

* `$std::axi4_lite_pkg` (axi_pkg.veryl) depends on `pkg::ADDR_W` (a.veryl).
* `$std::axi4_lite_if` (axi_if.veryl) depends on `$std::axi4_lite_pkg` (axi_pkg.veryl).
* `mdl` (a.veryl) depends on `$std::axi4_lite_if` (axi_if.veryl).

So Cyclic file dependency staill exists and this causes panic in [`type_dag`](https://github.com/veryl-lang/veryl/blob/e63206bbe82988198f07b2c6dc2ad901e06ad2e2/crates/analyzer/src/type_dag.rs#L460).
a.veryl -> axi_if.veryl -> axi_pkg.veryl -> a.veryl

PR #3284, opened 2026-08-25
fix(synthesizer): skip per-bit net allocation for inferred RAMs

### LINKED ISSUE #3232: High memory usage when synthesizing a large RAM-inferred array
## Summary

Synthesizing a large array uses much more memory than expected, even when the array is successfully inferred as RAM.

A `1,048,576 × 64-bit` memory uses about 10 GB of peak memory on my environment.

## Reproduction

```veryl
module memprobe (
    clk: input clock,
    we: input logic,
    addr: input logic<20>,
    wdata: input logic<64>,
    rdata: output logic<64>,
) {
    var mem: logic<64> [2 ** 20];

    always_ff {
        if we {
            mem[addr] = wdata;
        }
        rdata = mem[addr];
    }
}
```

```sh
/usr/bin/time -v veryl synth --top memprobe
```

Tested with:

```text
veryl 0.20.3-nightly (867deb1 2026-08-17)
```

Result:

```text
[synth-time] analyze (parse+pass1/2): 0.171s
[synth-time] build_gate_ir: 6.118s (0 cells, 64 ffs, 1 rams)
Maximum resident set size (kbytes): 10126140
```

The final result contains one RAM block:

```text
RAMbit ×67108864
```

## Suspected cause

RAM candidates are identified before variable allocation:

```rust
ctx.ram_vars = ram::infer_ram_vars(module, &ram);
ctx.allocate_variables(module)?;
```

However, `allocate_variables` still allocates `scalar_width * element_count` nets for every variable, including variables present in `ram_vars`:

```rust
let width = scalar_width * element_count;

let mut nets = Vec::with_capacity(width);
for bit in 0..width {
    nets.push(self.alloc_net(Some((name, bit))));
}
```

Later stages already special-case RAM variables:

- `preallocate_ff_cells` skips per-bit FF allocation for `ram_vars`
- RAM reads and writes use `RamCandidate`/`RamBuilder`
- the final IR contains an SRAM macro instead of FF and mux logic

Therefore, the initially allocated per-bit nets appear to be unused. Subsequent passes also allocate work arrays proportional to `module.nets.len()`, amplifying the peak memory usage.

Latte72R:
## Summary

Fixes #3232.

This change avoids allocating one `VarSlot` net for every stored bit of an inferred RAM.

Previously, `allocate_variables` allocated nets for the entire array even when the variable had already been recognized as RAM. 

Inferred RAM storage is represented by `RamBlock`, so this change leaves the corresponding `VarSlot` without per-bit nets. 
Other variables continue to use the existing allocation path.

## Regression Test

A new integration test synthesizes a 4096 × 64-bit memory and verifies that:

- exactly one RAM block is inferred;
- the number of nets originating from the memory variable does not scale with the memory depth.

Before the fix, the test failed because it found 262,144 variable-origin nets (`4096 × 64`). 
It passes after the per-bit allocation is skipped.

## Results

Using the reproducer and command from #3232, peak RSS in my environment decreased from 10,126,140 kB to 86,744 kB.


dalance:
Looks good!
I'll merge this PR.

Issue #3244, opened 2026-08-18
Proposal: floating-point types in std


Kenta11:
`std` has no floating-point support. `f32`/`f64` map to SystemVerilog `shortreal`/`real`, so they are simulation-only — synthesizable designs carrying IEEE 754 data have no shared type to agree on.

I would like to add one: the bit layout plus a few classification predicates (`is_zero`/`is_subnormal`/`is_inf`/`is_nan`).

**Is this wanted in `std`? And if so, how should it be shaped?**

A generic package is the obvious core:

```veryl
pub package fp_pkg::<EXP_W: p32, MTS_W: p32> {
    struct fp_t {
        sign: logic         ,
        exp : logic  <EXP_W>,
        mts : logic  <MTS_W>,
    }
    function is_nan (x: input fp_t) -> logic { ... }
}
```

but then callers repeat the widths everywhere:

```veryl
var a: $std::fp_pkg::<8, 23>::fp_t;
```

A caller can already work around that with one line of their own, which is a fair answer in itself:

```veryl
alias package fp32 = $std::fp_pkg::<8, 23>;

var a: fp32::fp_t;
let n: logic = fp32::is_nan(a);
```

The question is whether `std` should hand out the common names instead:

```veryl
pub package fp {
    // IEEE 754 floating-point
    type fp16  = fp_pkg::<5 , 10>::fp_t;
    type fp32  = fp_pkg::<8 , 23>::fp_t;
    type fp64  = fp_pkg::<11, 52>::fp_t;
    type fp128 = fp_pkg::<15, 112>::fp_t;
    // de facto standard floating-point
    type bf16 = fp_pkg::<8 , 7> ::fp_t;
}
```

That shortens the type but not the predicates:

```veryl
var a: $std::fp::fp32;                            // short
let n: logic = $std::fp_pkg::<8, 23>::is_nan(a);  // verbose
```

Is this the shape it should take? I am just as interested in implementations I have not thought of, or in a compiler change that would enable a better one. Inference is the obvious candidate: if `<8, 23>` could be recovered from the argument's type, the predicates would read `$std::fp::is_nan(a)`, which is what one actually wants — but generic inference currently solves only numeric parameters appearing in a port's width expression, so it cannot.

taichi-ishitani:
I think APIs (`is_none`) should be associated with the defined type (`fp_t`).
So how about implement the `impl` feature before adding `fp_pkg` and define them with this feature?
[...]

PR #3287, opened 2026-08-26
feat(veryl): Add method and associated const in Veryl

Kenta11:
This PR is discussed in #3244.

# Overview

In order to make coding easier, I introduce methods and associated consts in Veryl. A method is a function associated with a type. An associated const is a constant associated with a type.

```veryl
package fp_pkg {
    struct fp_t::<EXP_W: p32, MTS_W: p32> {
        sign: logic       ,
        exp : logic<EXP_W>,
        mts : logic<MTS_W>,
    }

    impl fp_t {
        const BIAS: u32 = (1 << (EXP_W - 1)) - 1;

        function is_nan (
            self,
        ) -> logic {
            return (self.exp == '1) && (self.mts != '0);
        }
    }

    // IEEE 754 floating-point
    type fp16  = fp_t::<5 , 10>;
    type fp32  = fp_t::<8 , 23>;
    type fp64  = fp_t::<11, 52>;
    type fp128 = fp_t::<15, 112>;
    // de facto standard floating-point
    type bf16 = fp_t::<8 , 7>;
}

module sample {
    var fp: fp_pkg::fp32;

    let is_nan: logic = fp.is_nan(); // the method is called
    let bias: u32 = fp_pkg::fp32::BIAS; // the associated const is read
}
```

# Supplement

By using a generic package, you can achieve similar things. However, the function call is verbose (see #3244 for details).

```veryl
// if using generic package
var a: fp_pkg::fp32;                           // var definition. This is short.
let n: logic = fp_pkg::<8, 23>::is_nan(a);  // function call. This is verbose.
```

# Restriction

Methods have the following restrictions in this implementation.

- The first parameter must be `self`, whose type is the target type of the `impl`. Its direction is implicitly `input`.
- The target type must be a struct, and the `impl` must be in the same scope as the target.
- A generic struct can be a target, but the `impl` applies to all of its instances. Specialization for a particular instance is not supported.
- An associated function, which has no `self`, is not supported.


taichi-ishitani:
@Kenta11 ,

> The target type must be a struct, and the impl must be in the same scope as the target.

Can you add `impl` support for other types (typedef, union, enum)?

Kenta11:
The other types can also be supported. However, I have not considered them yet, and additional changes are required. All of them are rejected for now.

I would like to open a separate issue for it.

taichi-ishitani:
> The other types can also be supported. However, I have not considered them yet, and additional changes are required. All of them are rejected for now.
> 
> I would like to open a separate issue for it.

Ok, I understood.

PR #3334, opened 2026-09-02
fix(analyzer): support an inout function argument instead of panicking

### LINKED ISSUE #3308: Analyzer panics on an `inout` function argument
## Overview

The analyzer panics when a function has an `inout` argument. The function does not need to be called, and its body can be empty.

This issue is reproducible with Veryl 0.20.3-nightly (dbd45f5 2026-08-29).

## Reproduction

```veryl
module main {
    function f (
        x: inout logic,
    ) {}
}
```

```console
$ RUST_BACKTRACE=1 veryl check
thread 'main' panicked at crates/analyzer/src/conv/declaration.rs:1228:34:
internal error: entered unreachable code
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: veryl_analyzer::conv::declaration::conv_function
   4: <() as veryl_analyzer::conv::Conv<...>>::conv
   5: veryl_analyzer::analyzer::Analyzer::analyze_pass2
   6: veryl::pipeline::analyze
   7: veryl::cmd_check::CmdCheck::exec
```

`veryl check` exits with status 101. Changing `inout` to either `input` or `output` avoids the panic.

The analyzer should handle the valid `inout` function argument without panicking.

ranaumarnadeem:
Closes #3308.

The analyzer panicked on any function with an `inout` argument, called or not:

```
thread 'main' panicked at crates/analyzer/src/conv/declaration.rs:1228:34:
internal error: entered unreachable code
```

Function arguments were mapped to a direction with only `VarKind::Input` and
`VarKind::Output` arms, so `inout` fell through to `unreachable!()`.

Adding the missing arm on its own wasn't enough though - it would have replaced a
loud crash with a quiet wrong answer, so this is three things:

- **The panic.** `VarKind::Inout` now maps to `Direction::Inout`.
- **The call site.** Both direction matches in `ir/function.rs` end in `_ => ()`, so
  an `inout` argument would have been silently dropped - the value never passed in,
  the result never written back. It now does copy-in / copy-out, pushing to both
  `inputs` and `outputs`.
- **`missing_tri`.** With the panic gone, that check fired instead. It's correct for
  a module's inout port, which is a net and needs `tri`, but a function's inout
  argument is a variable - your example uses `inout var [1:0] v` - and requiring
  `tri` made us emit `inout tri logic x`, which isn't valid in a function port list.
  The check is now scoped to non-function affiliation.

@taichi-ishitani's first example compiles and emits essentially the same SystemVerilog
he wrote by hand:

```systemverilog
function automatic void sum(
    inout logic [2-1:0] v
) ;
    v = v + 1;
endfunction

always_comb begin
    c = a;
    sum(c);
    b = c;
end
```

Tests: an analyzer test for that example, and a simulator test that checks the value
actually round-trips (10 in, 13 out) across every backend config rather than just
confirming it compiles. The existing `missing_tri` tests still pass, so module inout
ports are unaffected.

Not addressed here: the second comment, asking for an error when calling a function
with `output`/`inout` arguments inside `always_ff`. The DC output quoted there is
VER-134 on `b`, which that example assigns blockingly in reset (`b = 2'd0`) and
non-blockingly in the else branch (`b <= a`), so it looks like it's about `b` rather
than the `inc(a)` call. Happy to add it in a follow-up if needed


taichi-ishitani:
In SystemVerilog, functions cannot have ports of `net` type.
The code below causes a syntax error.

```systemverilog
module top;
  function automatic func(inout wire logic a);
  endfunction
endmodule
```

```console
Error-[SE] Syntax error
  Following verilog source has syntax error :
  	task/function port cannot be of net type
  "testbench.sv", 4: token is 'logic'
    function automatic func(inout wire logic a);
                                            ^

1 error
```

Therefore, I think a port witn `tri` type in a function should be reported as an error.
Have you implemented this kind of check?

taichi-ishitani:
> Not addressed here: the second comment, asking for an error when calling a function
> with output/inout arguments inside always_ff. The DC output quoted there is
> VER-134 on b, which that example assigns blockingly in reset (b = 2'd0) and
> non-blockingly in the else branch (b <= a), so it looks like it's about b rather
> than the inc(a) call. Happy to add it in a follow-up if needed

@tignear has implemented the checker for use of functions with side effect in `always_ff` blocks.
refs: #3229 
I think we should enhance the check or add a similar checker for this check.

PR #3346, opened 2026-09-04
fix(analyzer): keep the project qualification of subordinate generic args

taichi-ishitani:
### Problem

A generic instance whose arguments come from another project was emitted under a name nothing resolves, producing broken SystemVerilog while `veryl build` exits 0.

The failing shape needs three projects -- the root project, a middle dependency (`veryl_sample_b`), and a lower dependency (`veryl_sample_c`):

```veryl
// sample_c/src/sample_c_module.veryl
pub module sample_c_module::<WIDTH: u32> (
    i_d: input  logic<WIDTH>,
    o_d: output logic<WIDTH>,
) {
    always_comb {
        o_d = i_d;
    }
}
```

```veryl
// sample_b/src/sample_b_generic.veryl -- the import lives only in this file
import sample_c::sample_c_module;

pub module sample_b_generic::<WIDTH: u32> (
    i_d: input  logic<WIDTH>,
    o_d: output logic<WIDTH>,
) {
    inst u: sample_c_module::<WIDTH> (
        i_d: i_d,
        o_d: o_d,
    );
}
```

```veryl
// sample_b/src/sample_b_module.veryl -- a different file, no import
pub module sample_b_top (
    i_d: input  logic<sample_b_pkg::WIDTH>,
    o_d: output logic<sample_b_pkg::WIDTH>,
) {
    inst u: sample_b_generic::<sample_b_pkg::WIDTH> (
        i_d: i_d,
        o_d: o_d,
    );
}
```

`sample_c`'s emitted SystemVerilog loses the project prefix on the argument, and the argument then fails to resolve:

```systemverilog
module sample_c___sample_c_module__sample_b_pkg_WIDTH (   // `veryl_sample_b_` is missing
    input  var logic [-1:0] i_d,                          // WIDTH resolved to nothing
    output var logic [-1:0] o_d
);
```

It should be:

```systemverilog
module sample_c___sample_c_module__veryl_sample_b_sample_b_pkg_WIDTH (
    input  var logic [veryl_sample_b_sample_b_pkg::WIDTH-1:0] i_d,
    output var logic [veryl_sample_b_sample_b_pkg::WIDTH-1:0] o_d
);
```

This was found in a real multi-project design, where the same failure emitted a whole package whose function signatures, casts and comparison operands were empty -- the broken SystemVerilog surfaced only as a syntax error from the SystemVerilog compiler, long after `veryl build` had reported success.

### Cause

`insert_subordinate_generic_instances` qualifies each subordinate path against the **caller's** namespace, but the path is written inside `target`:

```rust
path.append_namespace_path(namespace, &target.namespace);
```

Two things go wrong there. With `target` = `sample_b_generic` and the path being `sample_c_module::<WIDTH>`:

- the cross-project test in `append_namespace_path` compares the caller's project against `target`'s project. Both are `veryl_sample_b`, so no prefix is added.
- the head resolution `resolve((&self.base_path(0), namespace))` fails, because `import sample_c::sample_c_module;` lives in `sample_b_generic.veryl` and is not visible from `sample_b_module.veryl`, where the caller is. The argument re-qualification loop at the end of `append_namespace_path` is never reached.

So the analyzer registers the instance as `__sample_c_module__sample_b_pkg_WIDTH`, while the emitter -- resolving from the scope where the reference is actually written -- asks for `__sample_c_module__veryl_sample_b_sample_b_pkg_WIDTH`:

```
[analyzer] name="__sample_c_module__sample_b_pkg_WIDTH"                   ns=veryl_sample_b::sample_b_generic
[emitter]  mapped="__sample_c_module__veryl_sample_b_sample_b_pkg_WIDTH"  result=Err(NotFound)
```

#### analyzer

Qualify the subordinate path from `target`'s inner namespace, which is where the reference is written. That namespace sees `target`'s imports and tells that the referred symbol lives in another project, so both the head and the arguments get their project prefix.

#### emitter

`scoped_identifier` silently emitted nothing when a resolvable path failed to resolve:

```rust
_ => {}
```

That arm can only be reached with `Err(_)` and `path.is_resolvable() == true` -- the symbol table and the emitted code disagree. Emitting nothing turns that into broken SystemVerilog that only surfaces as a syntax error from the SystemVerilog compiler, so it is now `unreachable!()`, matching what `emit_inst` already does for the same condition.

### Tests

`testcases/sample` becomes `testcases/sample_projects/sample_a`, and the reproduction is added as two new projects under the same directory:

- `sample_projects/sample_c` -- `sample_c_module::<WIDTH>`
- `sample_projects/sample_b` -- depends on `sample_c`; `sample_b_generic.veryl` holds the import and the generic module, `sample_b_module.veryl` holds the caller, `sample_b_pkg.veryl` holds `WIDTH`
- the root `Veryl.toml` gains `veryl_sample_b = {path = "testcases/sample_projects/sample_b"}`

All three conditions are required; dropping any of them makes the case pass:

- **two dependency levels below the root** -- with the caller in the root project, the root project prefix is applied at emission and masks the missing qualification
- **the import visible only in `sample_b_generic.veryl`** -- adding the same import to `sample_b_module.veryl` makes the head resolve, and the arguments are re-qualified
- **`sample_b_generic` and `sample_b_top` in different files** -- in one file they share the file scope, so the import is visible from the caller

Reverting the analyzer fix makes `make gen_sv` fail on `unreachable!()`. Emit order decides whether that is `emit_inst` (emitting `sample_b_generic.sv`) or `scoped_identifier` (emitting `sample_c_module.sv`); either way the build fails.


dalance:
Changing sample project path may cause .gitignore path change?

taichi-ishitani:
> Changing sample project path may cause .gitignore path change?

Yes. I updated `.gitignore` to adopt sample project change.

PR #3348, opened 2026-09-05
feat(analyzer): add the testbench attribute

### LINKED ISSUE #3272: Add `testbench` attribute
Testbench features such as `initial` statements and hierarchical references can only be used inside a module annotated with the `test` attribute.
However, there are cases where we want to use these features in a module which is not a test itself, e.g. a wrapper module to be instantiated in an SVTB.

For this purpose, add a `testbench` attribute. A module annotated with this attribute:

* can use testbench features inside it
* is not run as a test

ranaumarnadeem:
Closes #3272.

Adds a `#[testbench]` attribute. A module carrying it can use testbench features —
`initial` blocks, hierarchical references — but is not collected as a test, so it can
be instantiated from an SVTB as an ordinary module.

Everything gating testbench features already routed through a single flag,
`context.in_test_module`, derived from `ModuleProperty.test`. So the work was to make
that flag true for `#[testbench]` without registering the module as a test. No grammar
change was needed — attribute names are interned strings matched in `attribute.rs`,
not enumerated in `veryl.par`.

Four commits, each self-contained: parse the attribute, record it on the module symbol,
wire it into the gate, add a testcase.

Five places key on `ModuleProperty.test`. Two needed to change and three deliberately
did not, which is most of the design:

- `conv/ir.rs` — **changed.** Sets `in_test_module`, which ten call sites read.
- `reference_table.rs` — **changed.** Hierarchical references resolve through their own
  gate (`is_testbench_hier_ref`), not through `in_test_module`. I only found this because
  I checked behaviour after wiring the first one: `initial` worked and `dut.inner` still
  failed with `invisible_identifier`.
- `symbol_table::get_tests()` — **unchanged**, so a `#[testbench]` module is never
  collected to run. That's the "is not run as a test" half for free.
- `has_test_attribute` / `description_items_excluding_tests` — **unchanged**, so the
  module is *not* dropped from normal output.
- `emitter.rs` — **unchanged**, `TestType::Native` special-casing doesn't apply.

The last two are what make it usable from an SVTB. A `#[test]` module is emitted behind
`` `ifdef __veryl_test_* ``; a `#[testbench]` module is emitted unguarded:

```systemverilog
module veryl_testcase_Module93Tb;
    veryl_testcase_Module93 dut ();
    logic seen;
    initial begin
        seen = dut.inner;
    end
endmodule
```

Verified: the attribute accepts `initial` and hierarchical references; the same code
without it is still rejected; `#[testbench(foo)]` is rejected as taking no argument;
and `veryl test` on a project with both runs only the `#[test]` module. Tests cover
the first three, plus testcase 93 for the emitted output.

Three things I decided by judgment, and would rather confirm than assume:

1. The attribute takes no arguments. The issue doesn't mention any, and `#[test]`'s
   name/top arguments have no obvious meaning for something that isn't run.
2. `in_test_module` keeps its name even though it now means "testbench features are
   allowed" rather than "is a test". Renaming touches ten call sites and would bury
   the actual change; happy to do it separately if you'd prefer.
3. Nesting `#[test]` inside `#[testbench]` (or the reverse) is neither supported nor
   rejected — I didn't want to invent a rule. Worth deciding if it should error.


erniebnn:
I really like this attribute, it would simplify a lot of things. However, i wonder if it would be possible to make this work:

Given this code:

```sv
#[testbench]
module sram_wrapper (
    i_clk: input clock,
) {
    var w_enable: logic   ;
    var w_write : logic   ;
    var w_addr  : logic<2>;
    var w_wdata : logic<8>;
    var w_rdata : logic<8>;

    inst sram0: $comp::sram1rw #(
        DATA_WIDTH: 8,
        DEPTH: 4,
    ) (
        clk     : i_clk   ,
        i_enable: w_enable,
        i_write : w_write ,
        i_addr  : w_addr  ,
        i_wdata : w_wdata ,
        o_rdata : w_rdata ,
    );
}

#[test(tb_testbench_function)]
module tb_testbench_function {
    inst clk: $tb::clock_gen;

    inst dut: sram_wrapper ( i_clk: clk );

    initial {
        dut.sram0.load("file.bin");
        clk.next(10);
        $finish();
    }
}
```

trying to run it with `veryl test` throws this error:

```
[ERROR]       Failed to elaborate test (tb_testbench_function)
unsupported_description

  × unsupported description
    ╭─[/home/raupa/Development/verylbug/src/bug3.veryl:31:9]
 30 │     initial {
 31 │         dut.sram0.load("file.bin");
    ·         ─────────────┬─────────────
    ·                      ╰── this description is not supported by the simulator
 32 │         clk.next(10);
    ╰────
```

If the sram instance is in the same module, it works like a charm. Having this work would naturally integrate with the usage of the testbench attribute, because then you could have the code that loads a memory image in the main test module, while the sram itself sits in the `#[testbench]` module.

erniebnn:
Another bug i found, given this code:

```sv
#[testbench]
module m1 {
    var w_clk   : clock   ;
    var w_enable: logic   ;
    var w_write : logic   ;
    var w_addr  : logic<2>;
    var w_wdata : logic<8>;
    var w_rdata : logic<8>;

    inst sram0: $comp::sram1rw #(
        DATA_WIDTH: 8,
        DEPTH: 4,
    ) (
        clk     : w_clk   ,
        i_enable: w_enable,
        i_write : w_write ,
        i_addr  : w_addr  ,
        i_wdata : w_wdata ,
        o_rdata : w_rdata ,
    );

    initial {
        sram0.load("file.bin");
    }
}

#[test(test1)]
module test1 {
    var w_clk   : clock   ;
    var w_enable: logic   ;
    var w_write : logic   ;
    var w_addr  : logic<2>;
    var w_wdata : logic<8>;
    var w_rdata : logic<8>;

    inst sram0: $comp::sram1rw #(
        DATA_WIDTH: 8,
        DEPTH: 4,
    ) (
        clk     : w_clk   ,
        i_enable: w_enable,
        i_write : w_write ,
        i_addr  : w_addr  ,
        i_wdata : w_wdata ,
        o_rdata : w_rdata ,
    );

    initial {
        sram0.load("file.bin");
    }
}

#[test(test2)]
module test2 {
    inst dut: m1 ();

    initial {}
}
```

Both tests are effectively identical, only one uses the testbench attribute and one does not. `test1` succeeds normally, as expected. `test2` however fails with:

```[ERROR]       Failed test (test2): unknown component instance `sram0` ```

ranaumarnadeem:
Thank you. they're the same gap. The simulator's component registry is built only from the top test module, and the child-instantiation path drops a child's $comp instances on the floor. That was unreachable before, since the analyzer never allowed $comp outside the top module, and this PR only fixes the analyzer half.

$tb in a #[testbench] child does work end to end (I've added a test for it here), so the attribute stands on its own. $comp in a child needs the registry to collect from children and key by hierarchy, otherwise two instances of the same child collide, and dut.sram0.load() then falls out of the same naming. I'll do that as a follow-up on top of this.

Issue #3347, opened 2026-09-04
Veryl simulation panic with zero-width signal


Bapt-Roux:
## Describe the bug

A zero-width signal (`logic<0>`, i.e. a parameter-defaulted bus width
that's set to 0 to mean "unused") — a completely standard idiom for an
optional side-channel/tag port — crashes the native `veryl test`
simulator with an internal panic (`Option::unwrap()` on `None`), even
though `veryl check` and `veryl build` both accept the design fine and
emit standard, synthesizable SystemVerilog (`logic [-1:0]`, the usual
zero-width-vector convention every major simulator/synthesis tool
accepts).

## Minimal snippets to reproduce
```
module zero_width_repro #(
    param SIDE_W: u32 = 0,
) (
    i_clk : input  clock        ,
    i_rst : input  reset        ,
    i_data: input  logic<8>     ,
    i_side: input  logic<SIDE_W>,
    o_data: output logic<8>     ,
    o_side: output logic<SIDE_W>,
) {
    var data_r: logic<8>     ;
    var side_r: logic<SIDE_W>;

    always_ff (i_clk, i_rst) {
        if_reset {
            data_r = 0;
            side_r = 0;
        } else {
            data_r = i_data;
            side_r = i_side;
        }
    }

    assign o_data = data_r;
    assign o_side = side_r;
}

#[test(test_zero_width_repro)]
module test_zero_width_repro {
    inst clk: $tb::clock_gen;
    inst rst: $tb::reset_gen ( clk );

    var data_in : logic<8>;
    var side_in : logic<0>;
    var data_out: logic<8>;
    var side_out: logic<0>;

    inst dut: zero_width_repro #(
        SIDE_W: 0,
    ) (
        i_clk : clk     ,
        i_rst : rst     ,
        i_data: data_in ,
        i_side: side_in ,
        o_data: data_out,
        o_side: side_out,
    );

    initial {
        rst.assert();
        data_in = 8'h42;
        clk.next();
        $assert(data_out == 8'h42, "data path should pass through");
        $finish();
    }
}

```

Instantiate it with `SIDE_W: 0` (the default) inside a `#[test]` module
and run:

```
veryl check   # passes cleanly
veryl build   # passes cleanly, emits valid SV with logic [-1:0]
veryl test    # panics
```

## Expected behavior

The test runs normally (the design itself is trivial: a data path pass
through plus an always-zero-width side signal that's simply never
exercised).

## Actual behavior

[...]

PR #3356, opened 2026-09-07
feat: support a cast by a parenthesized expression

### LINKED ISSUE #3091: Cast by expression
SystemVerilog supports cast by expression if the given expression is evaluated as a constant positive integer.
Veryl should also support it.

Image

ranaumarnadeem:
Closes #3091.

Adds a parenthesized constant expression as a casting type, so the size form from the
LRM excerpt in the issue works:

```veryl
i_x as (W + 1)
(i_x - 2) as (Q + 1)
i_x as (W * 2 - 3)
```

emitting

```systemverilog
always_comb o_a = (W + 1)'(i_x);
always_comb o_b = (Q + 1)'((i_x - 2));
always_comb o_c = (W * 2 - 3)'(i_x);
```

Per your comment, the operand is wrapped with `()` rather than `<>`.

`P'(x - 2)` already worked — `i_x as W` emits `W'(i_x)`. Only the expression form was
missing, and it had no workaround: `as (W + 1)` was a parse error, while `as W + 1`
parses as `(i_x as W) + 1` because `as` binds to a `Factor`, silently emitting
`W'(i_x) + 1`.

Two commits: the feature, then the tests. They aren't split further because adding a
`CastingType` variant breaks every exhaustive match in the workspace, so grammar and
consumers can't be separated without leaving a commit that doesn't compile.

The generated parser is regenerated and committed as usual; most of that diff is
production renumbering in doc comments.

Existing cast forms are unaffected — `as W`, `as 8` and `as u32` emit exactly as
before — and a non-constant operand is already rejected (`i_x as (i_x)` gives
`invalid_operand`). No formatter change was needed.

One thing deliberately left out. The LRM sentence quoted in the issue also says it
shall be an error if the size is zero or negative, and this PR does not enforce that.
`as 0`, and `as Z` where `Z` is a parameter equal to 0, are already accepted today for
the existing literal and parameter forms, so rejecting only the parenthesized form
would be inconsistent, and rejecting all of them is a behaviour change wider than this
issue. It also overlaps the question you raised in #3347 about whether width-0
variables should be allowed at all. Happy to add the check across all cast forms once
that's settled, either here or separately.


crates/analyzer/src/conv/expression.rs
```diff
@@ -460,6 +460,18 @@ impl Conv<&CastingType> for ir::Factor {
 
                     return Ok(ir::Factor::Value(comptime));
                 }
+                CastingType::LParenExpressionRParen(x) => {
+                    let token: TokenRange = value.into();
+                    let (comptime, _) = eval_expr(context, None, x.expression.as_ref(), false)?;
```
dalance: How about using `eval_size` in utils.rs?
It includes type check, so case like `x as (u32)` can be detected.

```rust
let (comptime, _) = eval_size(context, x.expression.as_ref(), false)?;
```

PR #3368, opened 2026-09-08
Added as_slice to component Value

erniebnn:
I added an `as_slice` method to the Value struct, because there was no way to get the higher words (> 64 bits) of a `Value::Bits`. It mirrors the same functionality as the `Value::mask_xz` which already does the exact same thing for the XZ-Mask.
I am quite new to rust however, so please tell me if the name `as_slice` is too intrusive to common rust naming standards.


dalance:
Looks good! I think `as_slice` is appropriate.
I'll merge this PR.

PR #3376, opened 2026-09-09
Register index and select reads for FF classification

Kenta11:
`ff_opt` misclassifies a register when another `always_ff` block reads it only inside an array index or a bit select. The register looks unread, so `ff_opt` turns it into an immediate write. The reading block then sees the new value in the same clock edge.

```veryl
always_ff {
    if_reset { addr = 0; } else { addr = addr + 1; }
}
always_ff {
    q = mem[addr];
}
```

After the first edge, `q` takes `mem[1]` instead of `mem[0]`. `--disable-ff-opt` and external simulators are correct.


tignear:
Could you also consider covering the following patterns?

```veryl
 q = mem[$unsigned(addr)];
 q = $unsigned(mem[addr]);
```
Both should use the pre-edge value of addr when it is updated in another always_ff block on the same clock edge.

Kenta11:
> Could you also consider covering the following patterns?
> 
> ```
>  q = mem[$unsigned(addr)];
>  q = $unsigned(mem[addr]);
> ```
> 
> Both should use the pre-edge value of addr when it is updated in another always_ff block on the same clock edge.

Both patterns are fixed now, in the latest commit.

I also found HierVariable has the same gap. Its index and select aren't visited either. Something like q = dut.mem[idx] would hit the same bug when idx is a local register. I'll fix that here too.

PR #3416, opened 2026-09-15
fix(analyzer): check let and type identifiers against SystemVerilog keywords

### LINKED ISSUE #3397: `let` and `type` declarations skip the identifier check
# `let` and `type` declarations skip the identifier checks

**Version**: veryl 0.20.3

## Summary

`var x: logic;` is checked against the SystemVerilog keyword list.
But it seems `let x: logic = ...;` is not, and neither is `type x = ...;`.
Vivado picks these word as error.

## Sample

```veryl
module foo (
    i_a: input  logic,
    o_x: output logic,
) {
    let within: logic = i_a;
    type table = logic<8>;
    assign o_x = within;
}
```

ranaumarnadeem:
Closes #3397.

`var` identifiers are checked against the SystemVerilog keyword list, but `let` and
`type` identifiers are not, so this is accepted and then rejected by Vivado:

```veryl
let within: logic = i_a;
type table = logic<8>;
```

`check_identifier.rs` has a handler per declaration form, and `let` and `type` simply
had none, so they never reached the shared `check()` where `sv_keyword_usage` (and the
reserved `__` prefix check) fires. This adds the two handlers.

`let` maps to `Kind::Var`, so it also picks up the `prefix_var` / `suffix_var` / `case_var`
naming rules. That is a small behaviour change for anyone with those configured, but a
`let` is a variable and checking `var x` while skipping `let x` was the same
inconsistency as the keyword check. Flagging it in case you would rather keep `let`
out of the naming rules.

`type` gets a new `Kind::Typedef`, mapped to `None` in every naming match the same way
the `EnumItem` / `StructItem` / `UnionItem` kinds are, since there is no typedef naming
config. So a typedef gets the keyword and reserved-prefix checks only.

`let _ = ...` is skipped. The discard form is legitimate and the shared check flags an
anonymous identifier for every kind except `ClockDomain`; an existing test
(`mixed_struct_union_member`) relies on it, which is how I found out.

| | before | after |
|-|-|-|
| `var within` | `sv_keyword_usage` | unchanged |
| `let within` | accepted | `sv_keyword_usage` |
| `type table` | accepted | `sv_keyword_usage` |
| `let ok_name` / `type ok_t` | accepted | unchanged |
| `let _ = ...` | accepted | unchanged |


dalance:
handling `let` in block like below may be missing?

```systemverilog
always_comb {
    let within: logic = i_a;
    o_x = within;
}
```

ranaumarnadeem:
yea you are right it slipped through. I fixed it

PR #3426, opened 2026-09-17
Allow to use variables in range expression used for `case`/`inside`

### LINKED ISSUE #3425: `unevaluatable_value` error for range expression in `inside` operation
`unevaluatable_value` is reported for a range expression in `inside` operation.

```systemverilog
module a_module (
    i_a: input  logic<2>,
    i_b: input  logic<2>,
    i_c: input  logic<2>,
    o_d: output logic   ,
) {
    assign o_d = inside i_a { i_b ..= i_c };
}
```

```console
]8;;https://doc.veryl-lang.org/book/07_appendix/02_semantic_error.html#unevaluatable_value\unevaluatable_value (link)]8;;\

  💥 Value, which is not evaluable at elaboration, can't be used as case
  │ condition
  [Failed to read contents for label `Error location` (offset: 156, length: 11): OutOfBounds]
  help:
```

A scalar operand in `inside` operation is treated as an operand for `==?` operation.
To avoid mismatch between simulation and synthesis, this check is performed.
However, range expression is not so the check should not be performed for range expressions.

taichi-ishitani:
fix veryl-lang/veryl#3425


crates/analyzer/src/conv/utils.rs
```diff
@@ -3488,7 +3483,8 @@ fn range_item(
 
     let comptime = exp.eval_comptime(context, None);
     let lo_value = comptime.get_value().ok().and_then(|v| v.to_usize());
-    if !comptime.is_const {
+    if !comptime.is_const && range_item.range.range_opt.is_none() {
+        println!("error");
```
dalance: This seems to be debug code.

taichi-ishitani: I removed it.

